PR #2, opened 2017-07-25
NIFIREG-5: Initial implementation of nifi-registry-flow-data-model; i…

markap14:
…t will likely need significant updating before release time but this will get the ball rolling so that we can integrate with it on the NiFi side


bbende:
Looks good, going to merge, thanks!

PR #3, opened 2017-08-02
NIFIREG-6 Adding nifi-registry-bootstrap module

bbende:
- Enables registry to have the same setup as NiFi where a bootstrap process is started first which launches the main process
- All the standard operations such as start, stop, dump, etc, should now work from nifi-registry.sh
- Added placeholder LICENSE and NOTICE files to the assembly, need to fill in later
- Create nifi-registry-utils to hold utility code

### DIFF
--- a/nifi-registry-runtime/src/main/java/org/apache/nifi/registry/NiFiRegistry.java
+++ b/nifi-registry-runtime/src/main/java/org/apache/nifi/registry/NiFiRegistry.java
@@ -35,15 +35,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Random;
-import java.util.Timer;
-import java.util.TimerTask;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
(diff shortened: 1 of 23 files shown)


mcgilman:
Will review...

nifi-registry-runtime/src/main/java/org/apache/nifi/registry/NiFiRegistry.java
```diff
@@ -43,28 +90,242 @@ public void run() {
...
+            LOGGER.info("Jetty web server shutdown completed (nicely or otherwise).");
+        } catch (final Throwable t) {
+            LOGGER.warn("Problem occurred ensuring Jetty web server was properly terminated due to " + t);
+        }
+    }
+
+    /**
+     * Determine if the machine we're running on has timing issues.
+     */
+    private void detectTimingIssues() {
```
mcgilman: I'm not sure we need to detect timing issues for the registry. Can probably remove this check.

bbende:
@mcgilman thanks for reviewing, I made the updates you suggested, let me know if there is anything else.

mcgilman:
Thanks @bbende! This has been merged to master.

PR #4, opened 2017-08-08
NIFIREG-8: Registry Web API Stubs

kevdoran:
- Upgraded to latest version of Jersey
- Upgraded to io.swagger annotations, latest version
- Added resources for Bucket and Flow
- Extended model to have the beginnings of concept for Bucket

Next steps after this:
- Add API endpoints for user/group/policy management
- Possible rename some of the data model classes / packages to broaden the concept of "flow" data model to something more generic, as discussed with @bbende


bbende:
@kevdoran this looks great....

Running a full build I ran into an error when the swagger plugin was running, I had to change the config in the pom to wrap the locations and schemes in sub-elements like this:

```
<locations>
   <location>org.apache.nifi.registry.web.api</location>
 </locations>
 <schemes>
   <scheme>http,https</scheme>
 </schemes>
```
Not sure why I would run into that and you wouldn't. I'm using Maven 3.3.3, but the version of the plugin should be the same for both of us?

Other minor point, running with -Pcontrib-check there are some failures related to using wildcard imports. The current rules are just a copy from NiFi.

Anyway, I have the changes made for both of these if you want me to just include them when I merge it in, but I wanted to check about the plugin just to make sure I wouldn't be screwing up something else by changing that. Let me know what you think.

kevdoran:
@bbende Thanks for reviewing this!

That's odd regarding the swagger plugin. Thanks for doing a contrib-check, I did not realize that was setup for this sub-project yet. I'm good with you making those changes on merge. 

Thanks again!

PR #5, opened 2017-08-18
NIFIREG-7 Defining Provider API and initial filesystem providers

bbende:
This is a fairly large PR so let me give an overview of everything introduced here...

This PR introduces the concept of a "Provider" which will be an extension point to allowable plug-able implementations of each provider. Initially there are two types of providers:

- **MetadataProvider** - Responsible for tracking information about everything stored in the registry, such as what buckets there are, what objects are in those buckets, etc.
- **FlowPersistenceProvider** - Responsible for storing the contents of a versioned flow and all of its snapshots, this provider will be given the bytes of a snapshot and the metadata about it and will be responsible for persisting and retrieving those bytes

Initially the registry will provide an implementation of each provider that uses the local filesystem for storage/persistence:

- **FileSystemMetadataProvider** - Persists metadata to a configurable XML file using JAXB
- **FileSystemFlowPersistenceProvider** - Stores the bytes of a snapshot in a file, the general structure would be something like:
```
storage_dir/
    bucket_id/
        flow_id/
            snapshot_id/
                snapshot_id.flow 
```
The providers are configured through a providers.xml configuration file in the registry conf directory, such as:

```
<providers>
    <metadataProvider>
        <class>org.apache.nifi.registry.metadata.FileSystemMetadataProvider</class>
        <property name="Metadata File">./metadata.xml</property>
    </metadataProvider>
    <flowPersistenceProvider>
        <class>org.apache.nifi.registry.flow.FileSystemFlowPersistenceProvider</class>
        <property name="Flow Storage Directory">./flow_storage</property>
    </flowPersistenceProvider>
</providers>
```
Additional types of providers will likely be added as the scope of the registry expands to include other types of resources beyond just flows. There can be only one provider configured for each type of provider.

The providers are obtained through the ProviderFactory, currently accessed in NiFiRegistryResourceConfig to inject the providers into the REST resources.


nifi-registry-provider-api/src/main/java/org/apache/nifi/registry/flow/FlowProvider.java
```diff
@@ -0,0 +1,68 @@
...
+
+import org.apache.nifi.registry.provider.Provider;
+
+/**
+ * A service that can store and retrieve versioned flow snapshots.
+ *
+ * NOTE: Although this interface is intended to be an extension point, it is not yet considered stable and thus may
+ * change across releases until the registry matures.
+ */
+public interface FlowProvider extends Provider {
```
bbende: Kevin, thanks for taking a look... 

I originally had shortened the name from FlowPersistenceProvider to FlowProvider to make some of the supporting classes names less verbose, but with the way things turned out I think it is a non-issue now and we can go back to calling it FlowPersistenceProvider. I'll make that change and update the PR.

I envision there being a service above the providers, for now I'm calling it RegistryService, which would coordinate the operations... So for example, there would be a "snapshot method" on the service which would do something like the following:
-  Create any necessary versions or ids
-  Call the metadata provider to persist the snapshot metadata
-  Call a serializer to serialize the snapshot
-  Call the flow persistence provider to save the serialized bytes

For the search example, we probably have to think through what we want to support searching on. Searching by name is more like searching on metadata, where as if someone wants to find a snapshot that contains processor XYZ, then that would have to be searching into the content of each snapshot. I could potentially see having a "search provider" that separated search into its own thing, but need to think through it more.

kevdoran: @bbende, thanks for the reply. The example higher-level logic provided by the service layer makes sense to me. The type of search I have in mind for the not-to-distant-future is adding url-parameter filters to the rest endpoints that list resources, e.g.:

`/buckets?name=production`

I agree that search is a use case that would need further thought, and I see there are multiple ways that could be added given the foundation created in this PR. Thanks!

nifi-registry-utils/src/main/java/org/apache/nifi/registry/util/StringUtils.java
```diff
@@ -0,0 +1,99 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.registry.util;
+
+import java.util.Collection;
+
+public class StringUtils {
```
mcgilman: Unless I've missed something, we can probably use the Apache Commons `StringUtils` here. In NiFi this was avoided in certain cases to prevent loading it into the system classloader affecting all descendant classloaders.

bbende: Thanks @mcgilman I removed the StringUtils class and used the one from Apache Commons, let me know if there is anything else

mcgilman:
Thanks @bbende! This has been merged to master.

PR #6, opened 2017-08-24
NIFIREG-11 Data model refactoring

bbende:
Created BucketMetadata, FlowMetadata, and FlowSnapshotMetadata interfaces and their respective standard implementations, and updated the provider API to return these interfaces.

Also made slight refactorings to the standard data model:
- Bucket now has a Set of VersionedFlows
- BucketObject renamed to BucketItem
- Created VersionedFlowSnapshotMetadata to represent a snapshot without the contents
- Changed VersionedFlow to have a Set of VersionedFlowSnapshotMetadata


kevdoran:
reviewing...

nifi-registry-data-model/src/main/java/org/apache/nifi/registry/flow/VersionedFlowSnapshotMetadata.java
```diff
@@ -0,0 +1,117 @@
...
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+
+        final VersionedFlowSnapshotMetadata other = (VersionedFlowSnapshotMetadata) obj;
+
+        return Objects.equals(this.bucketIdentifier, other.bucketIdentifier)
+                && Objects.equals(this.flowIdentifier, other.flowIdentifier)
+                && Objects.equals(this.version, other.version);
+    }
```
kevdoran: This is a a nit-pick, but is including bucketIdentifier necessary for equals() and hashcode() here? It's omitted from BucketItem, where only the item's identifier is used. It should work fine both ways as we will be using UUIDs. I don't know if anything is needed, it just caught my eye as a minor inconsistency between BucketItem (and VersionedFlow which extends BucketItem) and VersionedFlowSnapshot/VersionedFlowSnapshotMetadata.

bbende:
@kevdoran thanks for reviewing!

I pushed up another commit that changes the equals method you mentioned... I agree that as long as flow id's are globally unique, then we only need to use flow id + version for the hash/equals of the snapshot method.

Regarding using the metadata objects in the REST layer... I think we should still leave the REST end-points using the original data model, and the service layer will translate from the provider API to the regular data model.

I also added a "BucketItemType" in BucketItem so that if the front-end was looking at bucket item they could easily look at field to determine the type.

If you are good with those changes then I am going to merge this in. Thanks!

kevdoran:
@bbende - I'm good with those changes, that all makes sense from my perspective as well. Code still builds for me as well so all looks good on my end.

PR #9, opened 2017-09-06
NIFIREG-16 Initial Spring integration for nifi-registry-web-api

bbende:
This PR converts the nifi-registry-web-api into a Spring Boot project using the spring-boot-jersey-starter to integrate Jersey endpoints with Spring's context. The nifi-registry-web-api is still packaged and deployed as a war as it was before.

In addition, dependencies for spring-core, spring-beans, and spring-context were added to the nifi-registry-framework module so that beans/methods can be appropriately annotated in the framework.


bbende:
Thanks Kevin, I'll go ahead and merge this.

PR #10, opened 2017-09-19
NIFIREG-18 Refactor MetadataProvider to use an embedded database

bbende:
- Initial plumbing for H2 database
- Setup Flyway with initial migration to define tables
- Setup entity classes with repositories
- Setup unit testing for repositories
- Removed existing MetadataProvider concept
- Removed provider impl module and moved remaining pieces into framework
- Added MetadataService with DatabaseMetadataService implementation
- Refactored RegistryService to use MetadataService
- Introduced verbose flag on some end-points to control loading nested objects
- Added ability to pass down paging/sorting params
- Added endpoints for available fields
- Adding ItemResource and ability to retrieve all items, or items by bucket
- Changing from Set to List on retrieval methods


nifi-registry-framework/src/main/java/org/apache/nifi/registry/db/repository/FlowRepository.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.List;
+
+/**
+ * Spring Data Repository for FlowEntity.
+ */
+public interface FlowRepository extends PagingAndSortingRepository<FlowEntity,String> {
+
+    @Query("SELECT f FROM FlowEntity f WHERE LOWER(f.name) = LOWER(:name)")
+    List<FlowEntity> findByName(@Param("name") String name);
```
kevdoran: Same as above regarding `@Query` annotation -> `findByNameIgnoreCase`.

nifi-registry-framework/src/main/java/org/apache/nifi/registry/db/DataSourceFactory.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.File;
+
+/**
+ * Overriding Spring Boot's normal automatic creation of a DataSource in order to use the properties
+ * from NiFiRegistryProperties rather than the standard application.properties/yaml.
+ */
+@Configuration
+public class DataSourceFactory {
+
+    private static final String DB_USERNAME_PASSWORD = "nifireg";
```
kevdoran: Is there any downside to having the username and password hardcoded? I suppose if an admin needs to secure the DB file they can use the OS files system permissions to prevent file access. Is there any other use case that a configurable user/pass would facilitate?

nifi-registry-web-api/src/main/java/org/apache/nifi/registry/NiFiRegistryApiApplication.java
```diff
@@ -14,27 +14,41 @@
...
     @Override
     protected SpringApplicationBuilder configure(SpringApplicationBuilder application) {
-        return application.sources(NiFiRegistryApiApplication.class);
+        final Properties fixedProps = new Properties();
+        fixedProps.setProperty("spring.jpa.hibernate.ddl-auto", "none");
+        fixedProps.setProperty("spring.jpa.hibernate.naming.physical-strategy", "org.hibernate.boot.model.naming.PhysicalNamingStrategyStandardImpl");
+
+        // turn on to debug
+        fixedProps.setProperty("spring.jpa.show-sql", "true");
+        fixedProps.setProperty("spring.h2.console.enabled", "true");
```
kevdoran: Did you intend to commit these debug props or was that included accidentally?

nifi-registry-resources/src/main/resources/conf/logback.xml
```diff
@@ -68,6 +68,11 @@
     
     <logger name="org.apache.nifi.registry" level="INFO"/>
 
+    <!-- To see SQL statements set this to DEBUG -->
+    <logger name="org.hibernate.SQL" level="INFO" />
+    <!-- To see the values in SQL statements set this to TRACE -->
+    <logger name="org.hibernate.type" level="INFO" />
+
```
kevdoran: When I ran the app. I noticed Hibernate log output going to bootstrap.log instead of app.log. Is that intended?

Example output in bootstrap.log:

```
2017-09-20 13:27:34,585 INFO [NiFi logging handler] org.apache.nifi.registry.StdOut Hibernate: select bucketenti0_.id as id1_0_0_, bucketenti0_.created as created2_0_0_, bucketenti0_.description as descript3_0_0_, bucketenti0_.name as name4_0_0_ from BUCKET bucketenti0_ where bucketenti0_.id=?
```

nifi-registry-framework/src/main/java/org/apache/nifi/registry/db/repository/BucketRepository.java
```diff
@@ -14,38 +14,21 @@
...
-     */
-    String getDescription();
+public interface BucketRepository extends PagingAndSortingRepository<BucketEntity,String> {
 
-    /**
-     * @return the metadata about the flows that are part of this bucket
-     */
-    Set<FlowMetadata> getFlowMetadata();
+    @Query("SELECT b FROM BucketEntity b WHERE LOWER(b.name) = LOWER(:name)")
+    List<BucketEntity> findByName(@Param("name") String name);
```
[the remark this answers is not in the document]
bbende: Great idea, I'll update to use that approach

nifi-registry-framework/src/main/java/org/apache/nifi/registry/db/repository/FlowRepository.java
bbende: Will update as above

nifi-registry-framework/src/main/java/org/apache/nifi/registry/db/DataSourceFactory.java
bbende: This is the same approach taken in NiFi so I went with it, the thinking is that the database can never be run in server mode and exposed outside of the file on the server, so at that point you are down to file permissions.

nifi-registry-web-api/src/main/java/org/apache/nifi/registry/NiFiRegistryApiApplication.java
bbende: The way I thought this was working was that even with those set to true, the logging didn't take place unless you also changed the levels in logback.xml as mentioned in your next comment, but apparently that is not true since you saw the output in bootstrap.log :)

nifi-registry-resources/src/main/resources/conf/logback.xml
bbende: Good find, will investigate

bbende:
@kevdoran I pushed a new commit that changed the repository methods to the spring-data naming convention, and I removed those two properties that were being set in the application class which seemed to eliminate the bootstrap logging, you can still enable the hibernate logging in logback.xml but it will only go to app log.

nifi-registry-framework/src/main/java/org/apache/nifi/registry/db/DataSourceFactory.java
kevdoran: Yep, that works for me. Just wanted to verify it.

kevdoran:
@bbende I reviewed your two new commits. Both look good to me. Thanks for making those fixes.

-Pcontrib-check is complaining about some whitespace on an empty line:

```
[INFO] --- maven-checkstyle-plugin:2.15:check (check-style) @ nifi-registry-web-api ---
[WARNING] src/main/java/org/apache/nifi/registry/NiFiRegistryApiApplication.java[44] (regexp) RegexpSinglelineJava: Line has trailing whitespace.
```

Other than that, LGTM.

bbende:
Thanks, I'll resolve the white space issue and merge to master

PR #13, opened 2017-09-28
NIFIREG-24 Adding RPM profile to assembly

bbende:


markap14:
Looks good. Was able to verify that it created the RPM and then was able to use that RPM to install the registry app. +1 merged to master. Thanks for knocking that out! Will be much easier for a lot of people.

PR #15, opened 2017-10-05
[NIFIREG-21] Leverage REST API and add unit test coverage and update …

scottyaslan:
….travis.yml

This PR integrates several of the new REST endpoints with the initial registry UI. With this the UI now supports the ability to view all flows for all buckets, to view all flows for a specific bucket, to view a specific flow in a specific bucket, to view all buckets, to create a new bucket, and to delete an existing bucket.


bbende:
+1 Was able to run the UI, created/deleted buckets, and viewed flows I created through REST API, will merge to master, thanks!

PR #20, opened 2017-10-12
NIFIREG-34 Add provider API to lib and cleaning up some other depende…

bbende:
…ncies


bbende:
Thanks Kevin, will merge to master

PR #21, opened 2017-10-13
[NIFIREG-31] update buckets data table to include search/filter capab…

scottyaslan:
…ilities and add bucket creation dialog


bbende:
+1 looks good, will merge to master

PR #22, opened 2017-10-17
NIFIREG-35 Initial NiFi Registry Client

bbende:
Adds a new module - nifi-registry-client - which uses Jersey client to interact with the REST API.

Refactored a couple of classes to the data-model project to be shared by client and framework.

I left this as two commits for now since the second commit adds exception handling, and in case we wanted to change that approach it would easier to go back to the previous commit.

Manual test class is in src/test/java - TestJerseyNiFiRegistryClient.


nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/FlowResource.java
```diff
@@ -0,0 +1,69 @@
...
+            value = "Retrieves the available field names that can be used for searching or sorting on flows.",
+            response = Fields.class
+    )
+    public Response getAvailableFlowFields() {
+        final Set<String> flowFields = registryService.getFlowFields();
+        final Fields fields = new Fields(flowFields);
+        return Response.status(Response.Status.OK).entity(fields).build();
+    }
+
+}
```
kevdoran: I've been thinking this over and am interested in getting your take on something:

It occurred to me that as we add additional types of BucketItem over times, each type need its own resource for looking up reference values, etc. That's not necessarily a problem, but it did prompt me to consider some alternatives. Perhaps combining all fields into a nested object returned by /fields (or, in that case, just extending the existing /buckets/fields to include nested objects holding the fields each bucket item type). I also wondered if we have a use case for these fields changing dynamically, and if not, should we just include a static list of valid field values in the documentation? There might be perfectly valid reasons for making the list dynamic and discoverable, so I'm not against it. I'm perfectly ok with adding this FlowResource for reference value look up as is. Just considering how well it will age over time and its utility.

pom.xml
```diff
@@ -39,12 +39,14 @@
...
         <module>nifi-registry-security-api-impl</module>
-        <module>nifi-registry-framework</module>
+        <module>nifi-registry-security-utils</module>
+	<module>nifi-registry-framework</module>
         <module>nifi-registry-provider-api</module>
         <module>nifi-registry-web-api</module>
         <module>nifi-registry-web-ui</module>
         <module>nifi-registry-bootstrap</module>
         <module>nifi-registry-assembly</module>
+	<module>nifi-registry-client</module>
```
kevdoran: Very minor, but if you end up making any other changes to this prior to merge, the indentation is off here and in the dependency added to nifi-registry-web-api's pom.xml.

nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/FlowResource.java
bbende: I like the idea of possibly combining them all into one call under /fields... the way it was currently done was simply to align with the resource type being returned by the REST end-point, but maybe that is not the easiest to maintain.

The reason I wanted to make it dynamic was because I envisioned someone later modifying the DB entities and having no idea that there were these end-points that returned field names, and now either fields are missing, or fields are present that were removed, etc.

A different option I considered was not having any of the /fields end-points and relying on the names in the entities returned from the REST API, but since the DB and REST layer are using different objects we would then need to maintain a mapping of field names between the two, which also seems error prone.

I'm kind of torn because it feels awkward right now to return entities with certain fields, and then say oh sorry these are the actual names you can search/sort on. Let me keep thinking about it.

bbende:
Thanks for the review! I am working on resolving the conflicts now.

bbende:
@kevdoran i rebased the code against master to resolve conflicts, i think for now we can leave the fields stuff and the interfaces alone and merge this in, and then maybe we can address either of them if needed, are you good with that?

kevdoran:
+1, I verified functionality after rebase. I'm good with that, thanks @bbende!

PR #23, opened 2017-10-18
NIFIREG-36 Add Json Mappers to server and client

kevdoran:
…odel classes

### DIFF
--- a/nifi-registry-client/src/main/java/org/apache/nifi/registry/client/impl/JerseyNiFiRegistryClient.java
+++ b/nifi-registry-client/src/main/java/org/apache/nifi/registry/client/impl/JerseyNiFiRegistryClient.java
@@ -16,6 +16,11 @@
  */
 package org.apache.nifi.registry.client.impl;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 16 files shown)


bbende:
@kevdoran thanks for taking care of this, what do you think about doing something like this?

https://github.com/apache/nifi/blob/master/nifi-commons/nifi-web-utils/src/main/java/org/apache/nifi/web/util/ObjectMapperResolver.java

You then register that as a provider with Jersey and I think whenever it marshalls JSON it will use that custom ObjectMapper.

I guess it depends if we want to make a global decision to never include nulls, but my preference would be to keep the Jackson annotations out of the domain objects if possible.

kevdoran:
Yeah, I agree that would be preferable. I'll push an updated that moves the Include rules to a custom ObjectMapper rather than introducing Jackson annotations on the domain objects themselves. Thanks for the feedback @bbende.

kevdoran:
This is now rebased against the branch that adds integration tests. We should wait to merge this until that branch is finalized and merged to master so this can be rebased again if needed, but let me know what you think of the changes in the last commit on this branch.

I added a couple other changes to this as well, including: 

- a custom object mapper in the client
- mapper providers are now loaded via package scanning in NiFiRegistryResourceConfig (possible by also adding Spring @Component annotations)

bbende:
+1 looks good, I like being able to find all the providers from scanning rather than having to explicitly list them, and good find on updating the client, I'll merge this to master

PR #24, opened 2017-10-18
NIFIREG-37: Add integration tests for web API to build

kevdoran:


nifi-registry-framework/src/main/java/org/apache/nifi/registry/db/DatabaseMetadataService.java
```diff
@@ -353,7 +353,9 @@ public void deleteFlow(final FlowEntity flow) {
 
     @Override
     public FlowSnapshotEntity createFlowSnapshot(final FlowSnapshotEntity flowSnapshot) {
-        flowSnapshot.setCreated(new Date());
+        if (flowSnapshot.getCreated() == null) {
+            flowSnapshot.setCreated(new Date());
+        }
```
kevdoran: @bbende, I know you are looking into a timestamp issue on NIFIREG-38

I added this because the time is also being set in the RegistryService on initial creation (eg, the POST .../versions code path) and returned without being updated by the data that was set here. So when the snapshot was fetched later (eg, GET .../versions/latest) the timestamp assertions I added in FlowsIT were failing.

If your pending solution for NIFIREG-38 makes this moot, I can drop this change from this PR.

bbende:
+1 looks good and awesome to have the automated testing, going to merge, thanks!

PR #26, opened 2017-10-20
[NIFIREG-41] increase disconnect tolerance and use trusty os, add tra…

scottyaslan:
…vis_wait 30


bbende:
Looks like this gets the travis build passing so I am a +1, we can just make a habit of running UI tests locally when reviewing, thanks @scottyaslan

PR #29, opened 2017-10-30
NIFIREG-33 Add LDAP and JWT auth support

kevdoran:
Summary: 
- Adds LdapIdentityProvider for authentication
- Adds /access/token endpoint for generating JWT for users that can authenticate with a configured IdenitiyProvider
- Adds JwtAuthenticationProvider for authentication
- Adds KeyService for key generation and tracking for signing JWTs
- Adds LdapUserGroupProvider for authorization
- Adds LDAP and JWT integration tests using Spring Boot embedded LDAP server
- Adds nifi-registry-client dependency to test scope of nifi-registry-web-api
- Refactors nifi-registry-security-api-impl into nifi-registry-framework
- Refactors all security related packages, such as o.a.n.r.authorization and o.a.n.r.authentication, under org.apache.nifi.registry.security

A lot of files are changed, but most are just import statements due to package renaming. For peer review, focus should probably be on: 
- KeyService and DatabaseKeyService implementation
- LdapUserGroupProvider, including XML configuration
- IdentityProvider interface and LdapIdentityProvider, including usage of ExtensionManager
- New project structure that rearranges security modules and packages, including impacted POMs


nifi-registry-framework/src/main/resources/META-INF/services/org.apache.nifi.registry.security.authorization.AccessPolicyProvider
```diff
@@ -0,0 +1,31 @@
+#
```
kevdoran: License got pasted twice, will fix

nifi-registry-framework/src/main/resources/db/migration/V1__Initial.sql
```diff
@@ -43,4 +43,11 @@ CREATE TABLE FLOW_SNAPSHOT (
     COMMENTS VARCHAR(4096),
     PRIMARY KEY (FLOW_ID, VERSION),
     FOREIGN KEY (FLOW_ID) REFERENCES FLOW(ID)
+);
+
+CREATE TABLE SIGNING_KEY (
+    ID VARCHAR2(50) NOT NULL,
+    TENANT_IDENTITY VARCHAR2(50) NOT NULL UNIQUE,
+    KEY_VALUE VARCHAR2(50) NOT NULL,
+    PRIMARY KEY (ID)
```
kevdoran: Let me know if the preference is to make this V2__*.sql. I added it here assuming that we didn't have to worry about any production deployments of nifi-registiry-0.0.1-SNAPSHOT 😄

bbende:
Reviewing...

bbende:
@kevdoran I looked into the issue with the app not shutting down cleanly... taking  a thread dump showed:

```
" (ldap-user-group-provider) - background sync thread" #28 prio=5 os_prio=31 tid=0x00007f86f06f2000 nid=0x140b waiting on condition [0x0000700002cf5000]
   java.lang.Thread.State: TIMED_WAITING (parking)
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x00000007a1560750> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
        at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.awaitNanos(AbstractQueuedSynchronizer.java:2078)
        at java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:1093)
        at java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.take(ScheduledThreadPoolExecutor.java:809)
        at java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1074)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1134)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```
Digging in further, I realized we don't have a concept of shutting down extensions, even though the authorizer, user/group provider, and policy providers do have a preDestruction method which would have shutdown that thread if we called it.

Tested a change where I made AuthorizerFactory extend DisposableBean and added:
```
@Override
    public void destroy() throws Exception {
        if (authorizers != null) {
            authorizers.entrySet().stream().forEach(e -> e.getValue().preDestruction());
        }

        if (accessPolicyProviders != null) {
            accessPolicyProviders.entrySet().stream().forEach(e -> e.getValue().preDestruction());
        }

        if (userGroupProviders != null) {
            userGroupProviders.entrySet().stream().forEach(e -> e.getValue().preDestruction());
        }
    }
```
This appears to work, so if you can add that for AuthorizerFactory and LoginIdentityProviderFactory then we should be good for that one.

nifi-registry-properties/src/main/java/org/apache/nifi/registry/properties/NiFiRegistryProperties.java
```diff
@@ -49,6 +49,8 @@
     public static final String SECURITY_NEED_CLIENT_AUTH = "nifi.registry.security.needClientAuth";
     public static final String SECURITY_AUTHORIZERS_CONFIGURATION_FILE = "nifi.registry.security.authorizers.configuration.file";
     public static final String SECURITY_AUTHORIZER = "nifi.registry.security.authorizer";
+    public static final String SECURITY_IDENTITY_PROVIDER_CONFIGURATION_FILE = "nifi.registry.security.identity.provider.configuration.file";
```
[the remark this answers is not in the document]
kevdoran: Yep, good catch. Will update this.

kevdoran:
@bbende thanks a lot for the thorough review! I fixed the integration test configuration so that they uncovered the runtime issues you ran into, and then fixed those. I made changes based on your other recommendations as well - thanks a lot for finding them and providing a solution for the clean shutdown of the authorizers. I agreed with all of those and added a commit with all changes. I rebased the branch off of the updated master as well.

Whenever you get a chance, you can re-review and let me know if there is anything else to be addressed. Thanks!

bbende:
This looks good, was able to get a JWT for a user from the test LDAP server and then use it to make another request.

Can we add the DisposableBean approach to LoginIdentityProviderFactory as well? Just to ensure we calling the preDestruction method for each identity provider.

After that I think this should be good to merge in, thanks!

kevdoran:
Yep, that makes sense. I updated my branch to add that.

Please squash everything into a single commit on merge. Thanks again!

bbende:
Looks good! Going to merge

PR #30, opened 2017-11-01
 NIFIREG-46 Add authorizedActions field to Bucket

kevdoran:
This builds upon PR #29 (NIFIREG-33), so only the commits on top of that need to considered.

The API now returns an additional field for a Bucket, telling the client what actions they are authorized to perform on that bucket. For example:

```
"authorizedActions": ["read", "write", "delete"]
```

With this change, `/buckets` is now a convenient initial endpoint to use to both check authentication of the client identity and discover buckets and authorizations available to the client.


kevdoran:
As PR #29 is wrapping up, I will wait until that is merged and then rebase this branch on the new master.

bbende:
+1 Verified the authorizedActions are returning correctly, will merge, thanks!

kevdoran:
Cool, thanks @bbende!

PR #31, opened 2017-11-03
NIFIREG-48: Added new nifi-registry-flow-diff module and updates to d…

markap14:
…ata model to accomodate what NiFi needs for integration.


bbende:
+1 Looks good, will merge to master

PR #32, opened 2017-11-03
[NIFIREG-40] Add coasters for notifications to the FDS NgModule. Also…

scottyaslan:
… upgraded to latest compatible versions of angular, angular material, covalent, and their dependencies


bbende:
+1 I rebased against master and resolved the conflict in the list viewer, everything looks good and verified the coasters display on deleting a bucket or flow

PR #33, opened 2017-11-03
NIFIREG-47 Improvements to nifi-registry-client

bbende:
- Using separate Jersey version for client only
- Added support for proxied entities
- Fixed deserializing of BucketItems
- Cleaned up methods with unnecessary parameters

### DIFF
--- a/nifi-registry-client/src/main/java/org/apache/nifi/registry/client/impl/JerseyBucketClient.java
+++ b/nifi-registry-client/src/main/java/org/apache/nifi/registry/client/impl/JerseyBucketClient.java
@@ -28,7 +28,9 @@ import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.MediaType;
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 /**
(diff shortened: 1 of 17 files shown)


bbende:
Not sure why one travis job failed, all tests passing for me locally

nifi-registry-client/src/main/java/org/apache/nifi/registry/client/impl/JerseyBucketClient.java
```diff
@@ -140,7 +150,8 @@ public Fields getFields() throws NiFiRegistryException, IOException {
                 target = target.queryParam("sort", sortParam.toString());
             }
 
-            return Arrays.asList(target.request().get(Bucket[].class));
+            return getRequestBuilder(target).get(List.class);
```
markap14: I think this should be done as above in the other getAll() call no? Getting a Bucket[] and then returning emptyList or Arrays.asList?

markap14:
@bbende code review all looks good except for the one inline comment I made. Thanks!

bbende:
@markap14 thanks for reviewing! pushed up a commit to address your comment, let me know if you are good and want me to merge

bbende:
This was merged by Mark, closing the PR

PR #35, opened 2017-11-08
NIFIREG-50 Adding additional metadata to some objects in the data mod…

bbende:
…el, changing serializers to only serialize VersionedProcessGroup


markap14:
@bbende  all looks good! +1 merged to master.

PR #36, opened 2017-11-10
NIFIREG-51: Fixed comparison between empty collection vs null and add…

markap14:
…ed notion of a DifferenceDescriptor to allow different verbiage for flow differences


bbende:
+1 looks good, will merge

PR #37, opened 2017-11-14
NIFIREG-45 Refactor LoginIdentityProvider interface as a more generic…

kevdoran:
… and flexible IdentityProvider 

The goals of this PR are to:
- unify standard identity provider implementations and the API for extensions to use the same interface (previously, standard implementations such as X509 and JWT relied on custom implementations
- allow for more flexible means for extensions to provide authentication other than username/password
- provide a standard way of bridging NiFi Registry IdentityProviders and Spring Security (see sequence diagram below)

identityproviderdesign

Please refer to https://issues.apache.org/jira/browse/NIFIREG-45 for more information.


bbende:
Nice work on this refactoring, seems like a good approach.

Noticed one minor thing when I was testing (and didn't really know what I was doing), I happened to POST to the new token/identity-provider endpoint, but I still had my URL form-encoding set so it was missing the Basic header, and I got a NPE:

```
Caused by: org.apache.nifi.registry.security.authentication.exception.IdentityAccessException: Unable to validate the supplied credentials. Please contact the system administrator.
	at org.apache.nifi.registry.security.ldap.LdapIdentityProvider.authenticate(LdapIdentityProvider.java:274) ~[nifi-registry-framework-0.0.1-SNAPSHOT.jar:0.0.1-SNAPSHOT]
	at org.apache.nifi.registry.web.api.AccessResource.createAccessTokenUsingIdentityProviderCredentials(AccessResource.java:220) ~[classes/:na]
	... 89 common frames omitted
Caused by: java.lang.NullPointerException: null
	at org.apache.nifi.registry.security.ldap.LdapIdentityProvider.authenticate(LdapIdentityProvider.java:238) ~[nifi-registry-framework-0.0.1-SNAPSHOT.jar:0.0.1-SNAPSHOT]
	... 90 common frames omitted
```
I'm guessing the call to extractCredentials returned null, but then not sure whats supposed to happen after that.

kevdoran:
Good catch! I added a null check that makes the LdapIdentityProvider implementation handle this gracefully. The next PR (#41) makes this even a bit more robust by adding a null check in the /access/token/identity-provider as well.

Thanks for the feedback, let me know if you see anything else to change prior to merging in. There are a couple deprecated classes (exception types) that I plan to address in a follow up PR as there were already a ton of classes touched here and I didn't want to introduce the noise of a bunch of classes with one-line changes to exception types in this PR.

bbende:
Thanks for update, looks good, will merge

PR #39, opened 2017-11-16
NIFIREG-55 Adding a startupFailure method to be called when exception…

bbende:
…s are caught in JettyServer which will called system.exit


bbende:
Thanks Kevin, going to merge..

PR #40, opened 2017-11-17
[NIFIREG-30] GET users/groups, delete user/group, delete users/groups…

scottyaslan:
…, create group, introduce group side nav


bbende:
+1 Verified I could access the UI with certs and SPNEGO, and was able to perform crud operations on users and groups, nice work! Will merge to master

PR #43, opened 2017-11-29
NIFIREG-58 Add current user permissions

kevdoran:
For authorizable resources, such as tenants and buckets and bucket items, add a permissions fields that indicates actions available to the current user such as read, write, delete.

Add a /registry/current-user endpoint to the Web API that shows the available action to top level resources, such as /buckets, /tenants, and /policies.

Add a configurable flag to tenants indicating if they are configurable or read-only based on the Authorizer & UserGroupProvider that is managing them.


kevdoran:
@scottyaslan, note this PR moves the loadCurrentUser endpoint from `/access` to `/registry/current-user`

bbende:
+1 verified the new functionality and everything looks good, will merge to master

PR #44, opened 2017-11-30
NIFIREG-59 Migrate Authorizer Exceptions

kevdoran:
Replaces Authorizer exception types with more general security provider exception types.


bbende:
+1 Looks good, will merge to master, thanks!

PR #45, opened 2017-11-30
NIFIREG-60 NiFi Proxy Identity Support

kevdoran:
Adds the ability to configure NiFi Identities to act as proxies for
FileAccessPolicyProvider in authorizers.xml

Note: I plan to add new test certs and integration test cases to SecureFileIT that tests this capability in conjunction with identity mapping and the X-ProxiedEntity client header. If it would be helpful, I can include those automated tests as part of this PR.

### DIFF
--- a/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/file/FileAccessPolicyProvider.java
+++ b/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/file/FileAccessPolicyProvider.java
@@ -34,6 +34,7 @@ import org.apache.nifi.registry.security.authorization.exception.UninheritableAu
 import org.apache.nifi.registry.security.authorization.file.generated.Authorizations;
 import org.apache.nifi.registry.security.authorization.file.generated.Policies;
 import org.apache.nifi.registry.security.authorization.file.generated.Policy;
+import org.apache.nifi.registry.security.authorization.resource.ResourceType;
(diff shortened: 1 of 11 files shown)


nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/file/FileAccessPolicyProvider.java
```diff
@@ -517,28 +522,22 @@ private void populateInitialAdmin(final Authorizations authorizations) {
...
-//                addUserToAccessPolicy(authorizations, ResourceType.Data.getValue() + ResourceType.ProcessGroup.getValue() + "/" + rootGroupId, node.getIdentifier(), WRITE_CODE);
-//            }
-//        }
-//    }
+    /**
+     * Creates a user for each NiFi client and gives each one write permission to /proxy.
+     *
+     * @param authorizations the overall authorizations
+     */
+    private void populateNodes(Authorizations authorizations) {
```
bbende: Looks like this needs to get called from the load method similar to populateInitialAdmin

kevdoran: Good catch, will update

bbende:
+1 all looks good now, was able to run the TestUserClient with a proxied entity based on setting up the initial NiFi identity in authorizers.xml, will merge

PR #46, opened 2017-12-04
NIFIREG-42 Adding IT test for NiFi Registry client and upgrading to S…

bbende:
…pring Boot 2.0.0.M7 so we can use Jersey 2.26


bbende:
Closing this PR in favor of PR 48 which includes this work.

PR #47, opened 2017-12-04
NIFIREG-62: Added ability to diff two Controller Services without dif…

markap14:
…f'ing entire flow

- Use HashSets instead of LinkedHashSets for VersionedProcessGroup, since there is no longer a need for the strict ordering


bbende:
+1 looks good, merging to master but will leave the JIRA open for any other changes

PR #49, opened 2017-12-06
NIFIREG-65 Return flow versions in reverse order

kevdoran:
Change the REST API endpoint that returns a list of all flow versions to sort by newest to oldest.


kevdoran:
@scottyaslan This changes the backend to return flow versions in the desired order for the front-end

bbende:
+1 will merge to master

PR #51, opened 2017-12-07
NIFIREG-61 Add support for encrypted config files

kevdoran:
Allows sensitive property values to be encrypted in the following configuration files: 

- nifi-registry.properties
- identity-providers.xml
- authorizers.xml 

A decryption key can be configured in bootstrap.conf that allows decrypting protected properties at runtime.


alopresto:
Will review.

alopresto:
Discussed with @kevdoran . He is going to make a small change so the `CryptoKeyHolder` doesn't maintain the master key value in memory for the lifetime of the application, but rather makes available to each servlet context the capability to retrieve the key from the `bootstrap.conf` file when necessary. Over the lifetime of the application, this value should be needed at most 3 times (NiFi Registry properties read, Identity provider read, and Authorizer read). 

He will also add some Javadoc and test cases.

kevdoran:
Had another followup conversation with @alopresto to confirm the desired behavior. We are on the same page now and I'm going to push one more update to this that simplifies the code path for loading the master key value from the configuration location. Please hold reviews/merge until that change is ready, which should be soon. @alopresto, I'll ping you when this is updated as we discussed. Thanks again for the help in getting this feature into Registry!

kevdoran:
@alopresto, I made the changes we discussed and I believe this branch is now ready to merge pending final review.

To @alopresto, @bbende, or whomever merges this: please do not squash on merge. The interim commit contains a useful history of the approaches @alopresto and I discussed, and the final commit removes an approach that may be useful to refer back to down the road should we ever revisit this.

alopresto:
I am reviewing this. As for squashing/merging, I too prefer the historical advantage of not squashing, but Apache NiFi policy is to perform the squash. I recommend creating a branch on your local fork which maintains this history in case it is needed, while the `master` branch will be more compact for community readability & maintenance (example -- [NIFI-3024-rebased-squashed](https://github.com/alopresto/nifi/tree/NIFI-3024-rebased-squashed) vs. [NIFI-3024](https://github.com/alopresto/nifi/tree/NIFI-3024)).

kevdoran:
Roger that, thanks. I'll maintain the history on my fork. It's rebased on current master, so feel free to squash all commits on merge. Or I can create a squashed branch if you prefer.

alopresto:
To be honest, if a "non-you" actor performs the merge, you don't have to worry about this. Your fork's branch will maintain history automatically. The master will be the squashed commit (usually cherry-picked onto the master branch).

nifi-registry-properties/src/main/java/org/apache/nifi/registry/security/crypto/CryptoKeyLoader.java
```diff
@@ -0,0 +1,87 @@
...
+                logger.error("Cannot read from bootstrap.conf file at {} to extract encryption key", bootstrapFile.getAbsolutePath());
+                throw new IOException("Cannot read from bootstrap.conf", e);
+            }
+        } else {
+            logger.error("Cannot read from bootstrap.conf file at {} to extract encryption key -- file is missing or permissions are incorrect", bootstrapFile.getAbsolutePath());
+            throw new IOException("Cannot read from bootstrap.conf");
+        }
+
+        if (CryptoKeyProvider.EMPTY_KEY.equals(keyValue)) {
+            logger.warn("No encryption key present in the bootstrap.conf file at {}", bootstrapFile.getAbsolutePath());
```
alopresto: I don't think this should be `WARN` level -- `INFO` is fine.

kevdoran:
Thanks @alopresto - updated that log level. Also rebased off the latest master. Let me know if you see anything else to be changed.

bbende:
Spoke with @kevdoran and was able to test the functionality in this PR using a pre-encrypted conf directory, verified the app started and worked successfully when encrypted values were present in nifi-registry.properties, identity-providers.xml, and authorizers.xml... will squash the commits and merge, thanks!

PR #52, opened 2017-12-11
NIFIREG-64 Updates Policy Authorization Behavior

kevdoran:
NIFIREG-64 Updates Policy Authorization Behavior

- Changes authorization to /policies to be based on /polices resource
- Changes per-bucket policies to inherit from base /buckets policy
- Cleans up framework classes related to Authorization


bbende:
+1 verified the new authorization, will merge, thanks!

PR #56, opened 2017-12-13
NIFIREG-71 Making JettyServer set wantClientAuth to true when needCli…

bbende:
…entAuth is false


kevdoran:
Hey @bbende, I tried this out and it worked for me. Using a NiFi Registry server I had configured for LDAP login authentication, I was able to access the server using a certificate trusted by the server without a token generated from login credentials, which was pretty cool.

In testing different scenarios (e.g., could I still authenticate with a JWT based on LDAP credentials without providing a client cert), it was difficult for me to change the behavior of what my browser was doing. Chrome still wanted to send the certificate I had previously selected from my Mac's System Keychain and ultimately I had to delete that certificate there to force it to prompt me again. Likewise, once I selected no certificate, I had to delete the server's cert that I had previously. This is quite

bbende:
Thanks! I'll go ahead and merge this

PR #60, opened 2017-12-15
NIFIREG-72 Fix API status codes for errors

kevdoran:
Fixes Tenants and Policies to return 404 for resource not found cases.


bbende:
Looks good, will merge

PR #64, opened 2017-12-21
NIFIREG-75 Fix User Groups for Composite Provider Config

kevdoran:
Removes user existence check from FileUserGroupProvider when group is created or updated. Replaces it with check in the Authorizer Decorator class created by Authorizer Factory, so that all providers are used. Also fixes bug when searching for group membership by user that returns results across all providers.

Also updates a package name in the authorizers.xml template.

UI fix for group membership deletion icon in user side panel.

### DIFF
--- a/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/authorization/UserGroup.java
+++ b/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/authorization/UserGroup.java
@@ -59,4 +59,13 @@ public class UserGroup extends Tenant {
         }
     }
 
+    public void addUser(Tenant user) {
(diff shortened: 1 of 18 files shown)


kevdoran:
Upon further discussion, may go with a different approach here. please hold off on merging this for now

kevdoran:
This has been updated with the latest approach

bbende:
Reviewing...

bbende:
+1 Ran through a full test with NiFi & Registry both configured with a composite user-group-provider, and put an LDAP user into a file-based group, and then made a bucket and gave the group access to that bucket, verified the LDAP user could use that bucket from NiFi... nice work

PR #68, opened 2017-12-22
[NIFIREG-83] deselect all users or groups in the add users to group a…

scottyaslan:
…nd add user to groups dialogs


bbende:
Verified the behavior and looks good, will merge

PR #74, opened 2017-12-30
NIFIREG-89 Add landing page for root URL '/'

dannylane:
Following on from a suggestion in the dev mailing list I created [NIFIREG-89](https://issues.apache.org/jira/browse/NIFIREG-89) to add a landing page informing the user they may have mistyped, similar to the page that currently exists in NiFi.
I followed a similar approach as was used in NiFi.

This is the page that has been added:
screenshot_landing_page

### DIFF
--- a/nifi-registry-assembly/pom.xml
+++ b/nifi-registry-assembly/pom.xml
@@ -118,6 +118,12 @@
         </dependency>
         <dependency>
             <groupId>org.apache.nifi.registry</groupId>
+            <artifactId>nifi-registry-web-error</artifactId>
+            <type>war</type>
+            <version>0.0.1-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.nifi.registry</groupId>
             <artifactId>nifi-registry-resources</artifactId>
             <classifier>resources</classifier>
             <scope>runtime</scope>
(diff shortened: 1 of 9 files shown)


bbende:
@dannylane thanks for the contribution! From a quick glance this looks good, will do a more detailed review when I get a chance.

One thing that will need to be done is providing a LICENSE & NOTICE inside the nifi-registry-web-errors WAR at src/main/resources/META-INF, similar to this:

https://github.com/apache/nifi-registry/tree/master/nifi-registry-web-ui/src/main/resources/META-INF

If the web-errors WAR does not bundle any third party libraries (JARs or JS/CSS) then the LICENSE will most likely be the regular ASLv2 license, and the NOTICE would just be:

```
nifi-registry-web-errors
Copyright 2014-2017 The Apache Software Foundation

This product includes software developed at
The Apache Software Foundation (http://www.apache.org/).

``` 
If the WAR does bundle other libraries then they may need to be mentioned in the license or notice.

dannylane:
Thanks @bbende, I've added the LICENSE & NOTICE. 
Let me know if there's anything else that needs to be changed.

alopresto:
There has actually been discussion on the NiFi side about removing this page. It is a legacy artifact that introduces a number of problems and doesn't serve a purpose that a silent redirect would not accomplish. Can we please hold off on merging this until the decision is made on the NiFi side about removing the page? Feel free to contribute to that discussion as well (I believe at this time, it has been conducted offline, so I will send an email to the dev mailing list to introduce the topic).

bbende:
@alopresto thanks for mentioning about the discussion of the NiFi landing page, I wasn't aware of that.

Even if we changed to a silent redirect, I _think_  we still want most of what is in this PR. What I mean is that this PR sets up the new WAR that handles the "/" context and currently displays the index.html from nifi-registry-web-errors WAR, which happens to be the landing page. 

If we wanted to redirect instead, I think we'd just be changing how the index.html worked at that point, maybe to something like index.jsp that called redirect? But in either case we need something to handle the root context of "/".

PR #78, opened 2018-01-08
[NIFIREG-78] disable new policy option when registry is unsecured

scottyaslan:


bbende:
Looks good, will merge

PR #81, opened 2018-01-09
NIFIREG-108 Change REST API buckets ordering

kevdoran:
Changes the list returned by GET /buckets to be alphabetically sorted
by bucket name. This is accomplished by adding an ORDER BY clause the
SQL in DatabaseMetadataService when querying buckets.


bbende:
Looks good, going to merge

PR #85, opened 2018-01-12
[NIFIREG-117] update sorting function name and add ellipsis to long t…

scottyaslan:
…ext in sidenav tables


scottyaslan:
Nice catch @kevdoran this is a def a bug. This bug also exists in when editing a user as well as when editing a bucket. I have created https://issues.apache.org/jira/browse/NIFIREG-124 to address this issue but I think this PR can be merged.

kevdoran:
Thanks for looking into it @scottyaslan. Will finalize my review and merge this if all else looks good.

kevdoran:
+1, merging to master

PR #86, opened 2018-01-16
NIFIREG-112 - Add a 'diff' endpoint to the API/Client for comparing 2 versions of a flow.

dannylane:
This PR is for an API endpoint to get a diff between 2 versions of a flow. 
[NIFIREG-77](https://issues.apache.org/jira/projects/NIFIREG/issues/NIFIREG-77) is the parent task and [NIFIREG-112](https://issues.apache.org/jira/projects/NIFIREG/issues/NIFIREG-112) is the sub task.
There is some discussion on the parent task around direction etc. 
Let me know if there's anything I should change. Thanks.


dannylane:
I should add the client change before I submit the PR

PR #91, opened 2018-01-18
NIFIREG-119 Detect incorrect authorizers config

kevdoran:
Adds stricter checks in AuthorizerFactory for unique ids within a
given type of provider and requires unique providers in composite
and composite-configurable user group providers. Failed validation
checks cause startup to fail. Adds test cases for these new rules.


nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/AuthorizerFactory.java
```diff
@@ -164,6 +169,9 @@ public Authorizer getAuthorizer() throws AuthorizerFactoryException {
 
                         // create each access policy provider
                         for (final org.apache.nifi.registry.security.authorization.generated.AccessPolicyProvider accessPolicyProvider : authorizerConfiguration.getAccessPolicyProvider()) {
+                            if (userGroupProviders.containsKey(accessPolicyProvider.getIdentifier())) {
```
kevdoran: used the wrong map here, will update

bbende:
Looks good, merging

PR #97, opened 2018-02-01
NIFIREG-134 Enable SpringBoot Actuator endpoints

kevdoran:
Enables Spring Boot 2 Actuator endpoints for remote system administration and monitoring:
https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#production-ready-endpoints 

Contains the following changes:
- Configures Jersey as a filter (previously was a servlet) that
  forwards requests to /actuator/* so they can be handled by Actuator
- Adds a ResourceAuthorizationFilter that performs authorization in
  the filter chain, and configures it to gate access to /actuator/*
- Adds test cases for ResourceAuthorizationFilter

For reviewers:
- Verify /nifi-registry-api/actuator is accessible in unsecured mode. It will return a listing of other endpoints (/nifi-registry-api/actuator/*) that you can also try.
- In secured mode, an initial admin should have access, as should any user who is granted access to the "/actuator" resource

Future enhancements (will create JIRAs if this is merged):
- Add Actuator usage instructions to Admin guide
- Add health checks that roll up into /actuator/health
- Add ActuatorClient to NiFiRegistryClient


bbende:
Tested this locally and verified the security of the actuator end-point, everything looks good to me, nice work! Will merge.

kevdoran:
Thanks for reviewing, @bbende! I'll open JIRAs for the follow on work that builds upon this capability to make it more useful for NiFi Registry users.

PR #98, opened 2018-02-01
NIFIREG-135 Fix versionCount for createFlowVersion result

kevdoran:


nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/RegistryService.java
```diff
@@ -589,8 +589,15 @@ public VersionedFlowSnapshot createFlowSnapshot(final VersionedFlowSnapshot flow
             // update the modified date on the flow
             metadataService.updateFlow(existingFlow);
 
+            // get the updated flow, we need to use "with counts" here so we can return this is a part of the response
+            final FlowEntity updatedFlow = metadataService.getFlowByIdWithSnapshotCounts(snapshotMetadata.getFlowIdentifier());
+            if (updatedFlow == null) {
+                throw new ResourceNotFoundException("Versioned flow does not exist for identifier " + snapshotMetadata.getFlowIdentifier());
+            }
+            final VersionedFlow updatedVersionedFlow = DataModelMapper.map(existingBucket, updatedFlow);
+
```
kevdoran: For the above block, I considered just modifying the in-scope versionedFlow object before returning, but thought it would be better (though more expensive) to just re-retrieve the flow, in case we add other fields to flow in the future that would be modified upon saving a new flow version.

Chaffelson: Sounds good.
Now if only I had a maven builder to Docker for NiFi-Registry...

bbende:
Looks good, will merge

bbende:
@kevdoran I messed up the magic commit message so this didn't auto-close, it is merged now if you can close it, thanks.

kevdoran:
Thanks for merging, @bbende!

PR #100, opened 2018-02-09
[NIFIREG-124] persist sidenav table sorting

scottyaslan:


kevdoran:
Reviewing...

scottyaslan:
Nice catch! I have updated this PR to address the issue with the "add to group" button. Please re-review. Thanks!

PR #102, opened 2018-02-15
"source" doesn't exist in regular /bin/sh

NKame:


NKame:
Thought it was not the real problem on my setup, but it was: reopening.

kevdoran:
Hi @NKame, thanks for the contribution. It seems the issue is the use of the [`source` function on line 88](https://github.com/apache/nifi-registry/blob/rel/nifi-registry-0.1.0/nifi-registry-resources/src/main/resources/bin/nifi-registry.sh#L88), which is a `bash` builtin.

While changing the shebang line to specify bash would fix this issue, I think we could also modify the script to remove any bash builtins and maintain sh/posix compatibility.

Could you try replacing line 88 in the script with this and see if that fixes the issue on your platform:

    . /etc/os-release

NKame:
I've tested the version with /bin/sh + . (instead of source): working on my setup, better than having a dependency towards bash.

PR #103, opened 2018-02-17
NIFIREG-146 REST API Documentation improvements

kevdoran:
Improves the REST API documentation and swagger spec, including:

- Corrects handling of collection response types in REST API docs
- Adds required access policy information in REST API docs
- Adds missing required=true tags to swagger spec
- Adds missing readOnly-true tags to swagger spec
- Adds security definitions to swagger spec
- Corrects VersionedConnection.zIndex field name in swagger spec

Functionality changes:

- Adds authorization check to the getFlowDiff endpoint


bbende:
Looks good, will merge

PR #106, opened 2018-03-15
[NIFIREG-153] inject angular router module into required components

scottyaslan:


bbende:
Looks good, will merge

PR #107, opened 2018-03-26
NIFIREG-157 Adding boolean to VersionedPropertyDescriptor to indicate…

bbende:
… if sensitive or not


kevdoran:
Looks good, will merge to master. Thanks for adding this @bbende

PR #99, opened 2018-02-02
[NIFIREG-126] adding some polish and testing around deep links

scottyaslan:
To test this PR try the following deep links:
Unsecured Registry:
-nifi-registry/explorer/grid-list/buckets/0 [should redirect to the nifi-registry/explorer/grid-list perspective and give an error dialog]
-nifi-registry/explorer/grid-list/buckets/0/flows/0 [should redirect to the nifi-registry/explorer/grid-list perspective and give an error dialog]
-nifi-registry/administration/users [should redirect to the workflow administration perspective and give an error dialog]
Secured Registry:
-nifi-registry/administration/workflow(sidenav:manage/bucket/000000000000) [should redirect to the workflow administration perspective and give an error dialog]
-nifi-registry/administration/users(sidenav:manage/user/000000000000) [should redirect to the workflow administration perspective and give an error dialog]
-nifi-registry/administration/users(sidenav:manage/group/000000000000) [should redirect to the workflow administration perspective and give an error dialog]

This PR also enables JS unit tests to run on travis ci.

PR #108, opened 2018-04-05
NIFIREG-158 Added ability to retrieve flow directly by id without kno…

bbende:
…wing the bucket

### DIFF
--- a/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/FlowResource.java
+++ b/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/FlowResource.java
@@ -18,19 +18,34 @@ package org.apache.nifi.registry.web.api;
 
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
 import io.swagger.annotations.Authorization;
+import io.swagger.annotations.Extension;
+import io.swagger.annotations.ExtensionProperty;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.nifi.registry.field.Fields;
+import org.apache.nifi.registry.flow.VersionedFlow;
+import org.apache.nifi.registry.flow.VersionedFlowSnapshot;
+import org.apache.nifi.registry.flow.VersionedFlowSnapshotMetadata;
+import org.apache.nifi.registry.security.authorization.RequestAction;
+import org.apache.nifi.registry.service.AuthorizationService;
 import org.apache.nifi.registry.service.RegistryService;
+import org.apache.nifi.registry.web.link.LinkService;
+import org.apache.nifi.registry.web.security.PermissionsService;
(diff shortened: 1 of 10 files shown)


kevdoran:
Will review...

bbende:
@kevdoran thanks, I have some more stuff to push up that I kept working on after submitting, incoming shortly...

kevdoran:
@bbende alright will wait until it's ready. thanks for letting me know!

bbende:
@kevdoran ok good to go now, ended up adding all of the GET operations from BucketFlowResource that could be done without a bucket id...

```
/flows/{flowId}
/flows/{flowId}/versions/{version}
/flows/{flowId}/versions/latest
/flows/{flowId}/versions/latest/metadata
/flows/{flowId}/versions
```

nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/FlowResource.java
```diff
@@ -62,4 +85,214 @@ public Response getAvailableFlowFields() {
...
+            final String flowId) {
+
+        final VersionedFlow flow = registryService.getFlow(flowId);
+
+        // this should never happen, but if somehow the back-end didn't populate the bucket id let's make sure the flow isn't returned
+        if (StringUtils.isBlank(flow.getBucketIdentifier())) {
+            throw new IllegalStateException("Unable to authorize access because bucket identifier is null or blank");
+        }
+
+        authorizeBucketAccess(RequestAction.READ, flow.getBucketIdentifier());
```
kevdoran: When I try this endpoint with an unauthorized user, I get the following response back:

```
HTTP/1.1 403 Forbidden
Connection: close
Date: Mon, 09 Apr 2018 14:08:27 GMT
Content-Type: text/plain
Content-Length: 101
Server: Jetty(9.4.3.v20170317)

Unable to view Bucket with ID 6eaeae9c-dbdb-4af3-a98e-4f3b880a0fb2. Contact the system administrator.
```

The 403 status code is good, but I'm not sure about the error message in the response. If someone is attempting to access the flow through /flows/{id}, I don't think the server should return the bucket id containing the flow, as that's leaking information the user would not otherwise have access to. It's a fairly harmless piece of information on it's own, but in a multi-tenant scenario it could reveal more than the owner of the bucket would like, especially if correlated with other information an attacker is able to obtain.

It's probably not a huge issue, but if you agree, we could strip this by wrapping the call to authorizeBucketAccess() in a try catch that obscures the error message returned in the response body. We would have to do this for all the GET methods in the FlowResource. Something like:

```
try {
    authorizeBucketAccess(RequestAction.READ, flow.getBucketIdentifier());
} catch (AccessDeniedException e) {
    throw new AccessDeniedException("User not authorized to view the specified flow");
}

By adding the root throwable to the cause of a custom exception, we could even keep the root cause with the bucket id in the logs for easier admin troubleshooting. 

Now that I think about it, that approach for customizing HTTP response error messages to differ from internal/logged error message probably be a better approach than the solution that we came up with in PR #99, which inadvertently introduced a "log & throw" pattern in order to maintain resource ids in the logs.

bbende:
@kevdoran good idea, I pushed up another commit that wraps the exception

PR #110, opened 2018-04-07
NIFIREG-160 - WIP - Hook provider

pvillard31:
To test it quickly, you can do the following changes:

./conf/providers.xml

````xml
<providers>

    <flowPersistenceProvider>
        <class>org.apache.nifi.registry.provider.flow.FileSystemFlowPersistenceProvider</class>
        <property name="Flow Storage Directory">./flow_storage</property>
    </flowPersistenceProvider>

    <flowHookProvider>
        <class>org.apache.nifi.registry.provider.hook.ScriptFlowHookProvider</class>
        <property name="Script Path">/tmp/test.sh</property>
        <property name="Working Directory"></property>
    </flowHookProvider>

</providers>
````

/tmp/test.sh

````bash
#!/bin/bash
echo $@ >> /tmp/registryHook.log
````

We can see this kind of ouput:

````
CREATE_BUCKET a50f3aac-a3b6-47ed-b0b0-eb7a1c0cb8b5
CREATE_VERSION 7091404e-1295-4d03-9053-06f2ecb83f91 707b31c8-6741-4525-9816-bd188f1f96eb 2 a comment anonymous
DELETE_BUCKET a50f3aac-a3b6-47ed-b0b0-eb7a1c0cb8b5
CREATE_BUCKET 95c7d15a-2ac8-494e-841a-c168e19c1c35
CREATE_BUCKET 2326a164-7ab3-45a3-af66-4adbf2cd4f32
DELETE_BUCKET 95c7d15a-2ac8-494e-841a-c168e19c1c35
DELETE_BUCKET 2326a164-7ab3-45a3-af66-4adbf2cd4f32
````

### DIFF
--- a/nifi-registry-framework/src/main/java/org/apache/nifi/registry/extension/ExtensionManager.java
+++ b/nifi-registry-framework/src/main/java/org/apache/nifi/registry/extension/ExtensionManager.java
@@ -22,6 +22,7 @@ import org.apache.nifi.registry.security.authorization.AccessPolicyProvider;
 import org.apache.nifi.registry.security.authorization.Authorizer;
 import org.apache.nifi.registry.security.authorization.UserGroupProvider;
 import org.apache.nifi.registry.flow.FlowPersistenceProvider;
+import org.apache.nifi.registry.hook.FlowHookProvider;
 import org.apache.nifi.registry.properties.NiFiRegistryProperties;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -56,6 +57,7 @@ public class ExtensionManager {
         classes.add(AccessPolicyProvider.class);
         classes.add(Authorizer.class);
         classes.add(IdentityProvider.class);
+        classes.add(FlowHookProvider.class);
         EXTENSION_CLASSES = Collections.unmodifiableList(classes);
     }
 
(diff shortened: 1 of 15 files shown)


bbende:
@pvillard31 I branched off your commit and resolved conflicts and took a stab at making the the event publishing/consuming asynchronous so that the main request path won't be impacted by any hang-ups from the providers. 

Also tried to make the event concept a little bit less specific to flows so that it can be easily re-used for any other types of items that we may store in the registry like extensions, assets, etc.

My branch is here: https://github.com/bbende/nifi-registry/commits/hook-provider

Let me know what you think about this approach. If we want to head down this path I can submit a PR to your branch, or I can open a new PR against registry that includes your commit + these changes. In the meantime I'll work on some unit tests.

pvillard31:
@bbende thanks for working on this! I really think it'll be a nice addition to the NiFi Registry and what you did looks great. I think it's best to close this PR and you submit a new one. It'll be easier, and I'll be happy to review it.

I'd just recommend keeping author and comment fields in the Event objects. Reason is: I'd imagine people that could add a specific tag in the comment to trigger automatic actions. Example, I have a flow I'm working on in Dev, and I consider the version I got is ready enough to get tested in Staging and deployed in production. I could commit my changes and comment with something like "[STAGING-READY]". This way, the deployment of the workflow in Staging would be automatically triggered.

I actually think it could be interesting to allow users overriding the comment when importing a new flow version from one registry to another. This would ease automatic deployment across multiple environments using the above mechanism. Just an idea... there could be something better. Do you have something in mind on your side?

PR #111, opened 2018-04-09
[NIFIREG-161] add vendor prefix to user-select style

scottyaslan:


pvillard31:
+1, merging, thanks for the quick fix @scottyaslan

PR #112, opened 2018-04-12
NIFIREG-162: Support Git backed PersistenceProvider

ijokarumawak:
- Added GitPersistenceProvider which manages Flow snapshot versions
by Git commits
- Updated FlowSnapshotContext to provide author
- Added new serialization data model version 2 which stores Flow
snapshots as JSON files


bbende:
Thanks @ijokarumawak ! Will review this soon.

bbende:
Been testing this and has been looking good so far...

I was doing a test where I configured to push to a remote, but I didn't supply a username/password because I honestly wasn't sure if the push would leverage my cached credentials at the OS level. 

So the push failed with an exception like:
```

Caused by: org.eclipse.jgit.errors.TransportException: https://github.com/bbende/nifi-versioned-flows.git: Authentication is required but no CredentialsProvider has been registered
	at org.eclipse.jgit.transport.TransportHttp.connect(TransportHttp.java:522) ~[org.eclipse.jgit-4.11.0.201803080745-r.jar:4.11.0.201803080745-r]
	at org.eclipse.jgit.transport.TransportHttp.openPush(TransportHttp.java:435) ~[org.eclipse.jgit-4.11.0.201803080745-r.jar:4.11.0.201803080745-r]
	at org.eclipse.jgit.transport.PushProcess.execute(PushProcess.java:160) ~[org.eclipse.jgit-4.11.0.201803080745-r.jar:4.11.0.201803080745-r]
	at org.eclipse.jgit.transport.Transport.push(Transport.java:1344) ~[org.eclipse.jgit-4.11.0.201803080745-r.jar:4.11.0.201803080745-r]
	at org.eclipse.jgit.api.PushCommand.call(PushCommand.java:169) ~[org.eclipse.jgit-4.11.0.201803080745-r.jar:4.11.0.201803080745-r]
	... 109 common frames omitted
```
This makes sense, but the result was that the flow was committed to the local repo, but because of the error when pushing, the error was thrown out of the REST layer and the response to NiFi indicated that starting version control failed. So it was left in a weird state where NiFi now thinks the process group is not under version control, but the local repo does have the first version saved.

Do you think we should catch any exceptions around the push inside the persistence provider and log them, but maybe not throw an error?

That would leave things in a more consistent state, but I'm kind of torn because then it may not be obvious to users that the pushes are failing unless they look in the registry logs.

bbende:
I noticed when I ran "git log" that the commit was made by "anonymous" which is correct since I was in an unsecure instance, but the email address of the commit ended up using the email from my ~/.gitconfig so I ended up with:

```
nifi-versioned-flows$ git log
commit 153690a2bd06d57ec416cb19b0582e2f7b138771 (HEAD -> master)
Author: anonymous <bbende@apache.org>
Date:   Thu Apr 26 10:58:40 2018 -0400

    Test
```
This is technically correct since that is the email address that should be found when there is not a more specific one, but should we make a property on the provider config like "Commit Email Address" ? or should we just leave it up to users to setup their gitconfig appropriately.

Ultimately we won't be able to have per-user email addresses anyway because when secured we will be using the identity of the proxied-entity as the author, and we have no way of knowing their email address.

nifi-registry-framework/src/main/java/org/apache/nifi/registry/provider/flow/git/GitFlowMetaData.java
```diff
@@ -0,0 +1,384 @@
...
+                if (bucketOpt.isPresent()) {
+                    bucket = bucketOpt.get();
+                } else {
+                    logger.debug("Bucket {} does not exist any longer. It may have been deleted.", bucketId);
+                    continue;
+                }
+            }
+
+            // E.g. DirA/DirB/DirC/bucket.yml -> DirC will be the bucket name.
+            final String[] pathNames = bucketFilePath.split("/");
```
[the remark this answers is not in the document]
ijokarumawak: Thanks @bbende for catching that. I added sanitization to convert invalid characters for filename such as '/', but left whitespaces as they are. Keep them as close to the original names but safe for file system names.

ijokarumawak:
@bbende I've moved push to different thread. Now it only commits locally on the web request handling thread. Used blocking queue to push fast enough but keep the thread waiting until new push is necessary, with this way no interval configuration is needed I think.

About the commit email address and author, I think using author and email address those are configured at Git repository as they are, and embedding NiFi registry end user name in a commit message would make more sense. By doing so, users can configure whatever they need at the git side, while being able to track how NiFi registry committed the change and by whom. How do you think?

bbende:
@ijokarumawak I agree about the email and author approach, after reading what you wrote in the admin guide, it makes sense.

Thanks for pushing those changes. Will try them out shortly.

nifi-registry-docs/src/main/asciidoc/administration-guide.adoc
```diff
@@ -895,3 +895,167 @@ Providing 2 total locations, including `nifi.registry.extension.dir.1`.
...
+
+Here is the data model version histories:
+
+|====
+|*Data model version*|*Since NiFi Registry*|*Description*
+|2|0.2|JSON formatted text file. The root object contains header and Flow content object.
+|1|0.1|Binary format having header bytes at the beginning followed by Flow content represented as XML.
+|====
+
+=== Migrating stored files between different Persistence Provider
```
bbende: I think instead of providing a tool we can just offer instructions for how to reset your registry to use the git provider, something like:
```
- Stop version control on all PGs in NiFi
- Stop registry
- Move the H2 DB and file-based flow dir somewhere for back up
- Configure git provider in providers.xml
- Start registry
- Recreate any buckets
- Start version control on all PGs again
```
This way the CLI doesn't need to depend on registry framework code since it is more of a client.

What do you think?

kevdoran: I agree it would be best to avoid a dependency on nifi-registry-framework for CLI if possible. @bbende, this approach you documented works, but loses the versioned PG history. Not sure how important that is, just mentioning it. 

I think the right analogy for this type of operation is a relational database migration. Most databases support this by giving you some way to export schema and data to a file in a portable SQL syntax that can then be imported into any compatible instance. This feature is useful for snapshotting, creating data backups, or migrating from one DB server to another. If we could support a similar feature in NiFi Registry, as export and import REST API endpoints, and corresponding CLI commands that would call the endpoints and write/read files, that would be a very powerful and flexible capability, both for backing up a NiFi Registry and also changing persistence providers.

This is probably an entirely new feature, so perhaps we could go with @bbende's documentation approach as a stop gap and open a JIRA for a more full-features export/import capability with no data loss.

bbende: Yea I agree that in the future we could probably offer a bulk import/export via the CLI which only needed to use the REST API. We already have a JIRA for some import/export end-points (https://issues.apache.org/jira/browse/NIFIREG-148), so it might just be some CLI commands to make use of those once they exist.

For this first release since we only have two providers, and since registry is very new, I think it is acceptable to say that if you want to use the git provider then you can start over with it.

nifi-registry-utils/src/main/java/org/apache/nifi/registry/util/FileUtils.java
```diff
@@ -389,4 +389,38 @@ public static void sleepQuietly(final long millis) {
             /* do nothing */
         }
     }
+
+
+    // The invalid character list is copied from this Stackoverflow page.
+    // https://stackoverflow.com/questions/1155107/is-there-a-cross-platform-java-method-to-remove-filename-special-chars
+    private final static int[] INVALID_CHARS = {34, 60, 62, 124, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
```
bbende: Should we add a space to this list? 

Technically a space is valid, but currently a bucket name of `This / is a / test` becomes `This _ is a _ test` but I was originally thinking it would be `This_is_a_test` or I guess technically it would have multiple underscores like `This___is_a___test` ?

bbende:
Async pushing looks good, so I think final two items are updating the migration guidance and whether or not we want to leave spaces in the bucket/flow names.

alopresto:
I vote for dropping spaces from the filename. Yes, it's technically valid, but it's escaped differently on different OSes and will just cause problems. There is no harm in removing it.

ijokarumawak:
@bbende I've addressed the final two items, updating migration guide and added white space to invalid filename characters. Also, updated how NiFi Registry username is used in Git commits. Please check the additional commits. Thanks!

bbende:
@ijokarumawak thanks for the updates, everything looks good, going to merge... thanks for all the work on this!

kevdoran:
Nice work @ijokarumawak and everyone who contributed to this. This is a major feature to have as part of NiFi Registry and will certainly be useful.

PR #113, opened 2018-04-17
NIFIREG-163 Added JDK information to System Requirements section of A…

andrewmlim:
…dmin Guide

Also added formatting to a default system property to be consistent with others.


scottyaslan:
Reviewing...

PR #117, opened 2018-05-09
NIFIREG-160 Implement a hook provider

bbende:
For whoever reviews/merges this, please keep the commit history and don't squash.


bbende:
For an easy way to test this you can turn uncomment the LoggingEventHookProvider in providers.xml and then use the registry as normal to create buckets and save flows from NiFi, then tail or inspect logs/nifi-registry-event.log

bbende:
@pvillard31 this should be good to go pending any review feedback, let me know if you have any cycles to take a look, thanks

pvillard31:
Hi @bbende - I should be able to review and test shortly, will let you know. Thanks again for this PR!

nifi-registry-framework/src/main/java/org/apache/nifi/registry/provider/flow/StandardFlowSnapshotContext.java
```diff
@@ -33,8 +33,8 @@
     private final String flowName;
     private final int version;
     private final String comments;
-    private final long snapshotTimestamp;
```
pvillard31: Really nitpick but I think the changes on this class seem unnecessary, no?

nifi-registry-framework/src/test/java/org/apache/nifi/registry/provider/hook/TestScriptEventHookProvider.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.nifi.registry.provider.ProviderCreationException;
+import org.apache.nifi.registry.provider.ProviderFactory;
+import org.apache.nifi.registry.provider.StandardProviderFactory;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class TestScriptEventHookProvider {
+
+    @Test(expected = ProviderCreationException.class)
+    public void testBadScriptProvider() {
```
pvillard31: can't test on my side but is it going to work for a build on Windows?

pvillard31: ok nevermind, since we expect it to fail, it should also fail on Windows

nifi-registry-framework/src/test/java/org/apache/nifi/registry/service/TestRegistryService.java
```diff
@@ -81,6 +81,7 @@
     public void setup() {
         metadataService = mock(MetadataService.class);
         flowPersistenceProvider = mock(FlowPersistenceProvider.class);
+        // eventHookProvider = mock(EventHookProvider.class);
```
pvillard31: To be removed?

pom.xml
```diff
@@ -110,8 +110,8 @@
         <jax.rs.api.version>2.1</jax.rs.api.version>
         <jersey.version>2.26</jersey.version>
         <jackson.version>2.9.2</jackson.version>
-        <spring.boot.version>2.0.0.M7</spring.boot.version>
-        <spring.security.version>5.0.0.RELEASE</spring.security.version>
+        <spring.boot.version>2.0.2.RELEASE</spring.boot.version>
```
pvillard31: It seems we are also specifying the versions in the notice files. Should be updated as well.

pvillard31:
@bbende  I just did some tests with various possible configurations and different hooks, it's working as expected. The dedicated log file is really great and useful. Let me know if you want to do something regarding my comments or I can take care of it while merging if you are OK with it.

bbende:
@pvillard31 thanks for the review, i'm fine with you doing the minor cleanup on merge, thanks!

pvillard31:
Merged to master, thanks!

PR #120, opened 2018-05-31
NIFIREG-56: Added element ids to facilitate bucket testing use cases

andrewmlim:
These are the ids I needed to do some automated UI testing around Bucket test cases (Bucket creation/deletion/duplicates etc.).  If more ids are needed for other test cases, will file separate more specific Jiras.


scottyaslan:
Will review...

PR #123, opened 2018-06-07
[NIFIREG-39] leverage npm @nifi-fds/core

scottyaslan:


scottyaslan:
Please do not merge.

mcgilman:
@scottyaslan Thanks for the PR! It appears that some of the font weights have changed. I'm not sure if we want to try to retain the existing look or use the new styles from the newer versions of angular-material. Functionality seems consistent but font-weights and even some spacing appear slightly different.

PR #124, opened 2018-06-13
NIFIREG-174 Fixing start-up to look for the system properties specify…

bbende:
…ing the location of properties and bootstrap, and fallback to relative paths


kevdoran:
reviewing...

PR #132, opened 2018-07-24
Incorrect reference to providers.xml

colinfindlay-nz:


kevdoran:
Thanks for the contribution @colinfindlay-nz! There was another PR with the same set of changes. Unfortunately can only merge one, so I closed this PR and added you as a co-author on the other commit with the same changes.

PR #133, opened 2018-08-08
NIFIREG-190

jdye64:


jdye64:
Looks like the Travis CI build is failing trying to extract the npm binary for building the web ui section. This doesn't have anything to do with the PR so don't believe the build is truly failing here...

pvillard31:
Had a quick look through the code and it LGTM, will try to give it a try today or tomorrow if no else has the cycles to do so.

kevdoran:
Will review...

jdye64:
I had contemplated placing the logic that checked if the hook provider should be invoked for an event or not right above https://github.com/apache/nifi-registry/blob/7e4e9713406f3bb3ac1544b6ad69fbdd437f794c/nifi-registry-framework/src/main/java/org/apache/nifi/registry/event/EventService.java#L73 but decided to place that logic in each hook provider in case someone wanted to write their own custom hook provider and want to apply different white listing logic that would allow them to circumvent such filtering. How do you both feel about that?

kevdoran:
If we are allowing each implementation to decide the logic in the handle method, I don't see the value in the "share property" concept as documented:

> There are certain properties that are shared amongst all of the NiFi Registry provided Event Hook implementations.

I don't think we should introduce the concept of a shared property that is documented as being honored by all implementations when we really have no way to guarantee that. 

In this PR, the implementation that does handle the Event Whitelist is implemented as an Abstract base class in the framework, which would make it difficult for third-party extension providers to implement (or they could ignore it if they want as it is not part of the API).

If we want to make it universal, I think we would need to introduce a dedicated field for whitelist that is enforced by the framework code (perhaps using a decorator / wrapper), not the extension. Otherwise, don't think we make it universal at all, it can just be an optional thing that third parties can choose to implement. 

One way to make it more obvious to custom implementations that this is intended to be implemented would be to make it part of the interface they have to implement (otherwise it is just documentation and hard to enforce). For example, we could add a new method with a default implementation to the `EventHookProvider` interface. For example:

```
    default boolean shouldHandle(EventType eventType) {
        return true;
    }
```

If it's not overridden with a custom implementation, the event hook provider will always be called. If it is overridden, they can return a boolean for the event types they are interested in, either as a hard-coded set or a user-configurable set (they would still need to document their own configuration property for that).

nifi-registry-docs/src/main/asciidoc/administration-guide.adoc
```diff
@@ -1094,3 +1094,63 @@ Here is the data model version histories:
...
+
+....
+<eventHookProvider>
+    <class>
+      org.apache.nifi.registry.provider.hook.ScriptEventHookProvider
+    </class>
+    <property name="Script Path"></property>
+    <property name="Working Directory"></property>
+    <!-- optional -->
+    <property name="Event Whitelist">CREATE_FLOW,UPDATE_FLOW</property>
```
kevdoran: Comma-separated lists in XML is inconsistent with how we handle configuring lists in other contexts. For example, the initial user identities in Authorizers.xml:

```
    <!--
        The FileUserGroupProvider will provide support for managing users and groups which is backed by a file
        on the local file system.

        - Users File - The file where the FileUserGroupProvider will store users and groups.

        - Initial User Identity [unique key] - The identity of a users and systems to seed the Users File. The name of
            each property must be unique, for example: "Initial User Identity A", "Initial User Identity B",
            "Initial User Identity C" or "Initial User Identity 1", "Initial User Identity 2", "Initial User Identity 3"

            NOTE: Any identity mapping rules specified in nifi-registry.properties will also be applied to the user identities,
            so the values should be the unmapped identities (i.e. full DN from a certificate).
    -->
    <userGroupProvider>
        <identifier>file-user-group-provider</identifier>
        <class>org.apache.nifi.registry.security.authorization.file.FileUserGroupProvider</class>
        <property name="Users File">./conf/users.xml</property>
        <property name="Initial User Identity 1"></property>
    </userGroupProvider>
```

We may want to reconsider in order to keep this consistent....

nifi-registry-framework/src/main/java/org/apache/nifi/registry/provider/hook/AbstractHookProvider.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Determines if the Event should be handled or ignored based on the user configured Event Whitelist property
+     *
+     * @param event
+     *  Event that has been passed to the provider by the framework.
+     *
+     * @return
+     *  True if the event should be handled by this provider and false otherwise.
+     */
+    public boolean handleEvent(Event event) {
```
kevdoran: As commented on the PR, this should perhaps be implemented differently. But if going with the Abstract base class approach, I would change the signature of this method to:

```
protected boolean shouldHandle(Event event)
// or
protected boolean shouldHandle(EventType eventType)
```

nifi-registry-framework/src/main/java/org/apache/nifi/registry/provider/hook/AbstractHookProvider.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public void handle(Event event) throws EventHookException {
+        // Abstract event handling logic can be implemented here.
+    }
+
+    @Override
+    public void onConfigured(ProviderConfigurationContext configurationContext) throws ProviderCreationException {
+
+    }
```
kevdoran: This is an abstract class so we don't need the empty interface methods for `handle(...)` and `onConfigured(...)`

nifi-registry-framework/src/main/java/org/apache/nifi/registry/provider/hook/LoggingEventHookProvider.java
```diff
@@ -36,6 +36,10 @@ public void onConfigured(final ProviderConfigurationContext configurationContext
 
     @Override
     public void handle(final Event event) throws EventHookException {
+
+        // Purposely leaving off "handleEvent" logic as its assumed that everything should be logged.
+        //handleEvent(event)
```
kevdoran: but that contradicts the documentation of the shared property... again I think we need to discuss what the intention is and then implement the code to match

nifi-registry-framework/src/main/java/org/apache/nifi/registry/provider/hook/AbstractHookProvider.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * An AbstractHookProvider class that serves as a place to keep common Event Hook implementation details.
+ */
+public abstract class AbstractHookProvider
+        implements EventHookProvider {
+
+    static final Logger LOGGER = LoggerFactory.getLogger(AbstractHookProvider.class);
+
+    static final String EVENT_WHITELIST = "Event Whitelist";
+    protected List<EventType> whiteListEvents = null;
```
kevdoran: This would be better as `Set<EventType>` so that `contains()` lookups are O(1)

jdye64:
Ok. After reading through all of these I'm thinking the best way to handle this and have it operate the way I had envisioned is to take the approach of adding the handleEvent() method in the EventProvider interface as outlined above. I'm going to start on that now and just remove the Abstract class all together.

kevdoran:
Thanks @jdye64! That sounds good, I'll give it another look when those changes are ready.

nifi-registry-provider-api/src/main/java/org/apache/nifi/registry/hook/EventHookProvider.java
```diff
@@ -36,4 +36,18 @@
      */
     void handle(Event event) throws EventHookException;
 
+    /**
+     * Examines the values from the 'Event Whitelist X' properties in the hook provider definition to determine
+     * if the Event should be invoked for this particular EventType
```
kevdoran: 'Event Whitelist X' property has been renamed in this commit to 'Whitelisted Event Type '.

nifi-registry-framework/src/main/java/org/apache/nifi/registry/provider/hook/AbstractEventHookProvider.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.nifi.registry.provider.ProviderCreationException;
+import org.springframework.util.CollectionUtils;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public abstract class AbstractEventHookProvider
```
bbende: This should be part of the provider api module right? If someone implements a custom event hook they shouldn't have to create a dependency on nifi-registry-framework.

kevdoran:
A couple of other things I thought of:

- The default [providers.xml](https://github.com/apache/nifi-registry/blob/7e4e9713406f3bb3ac1544b6ad69fbdd437f794c/nifi-registry-resources/src/main/resources/conf/providers.xml#L34) should maybe be updated to include placeholders/explanation for the whitelist properties.

- By adding the `shouldHandle(EventType)` method to the EventHookProvider interface, should we move the conditional execution of the hook's `handle(Event)` method to a check where it is called in [EventService](https://github.com/apache/nifi-registry/blob/7e4e9713406f3bb3ac1544b6ad69fbdd437f794c/nifi-registry-framework/src/main/java/org/apache/nifi/registry/event/EventService.java#L73)?

jdye64:
I think if the EventType is null, which I'm like you I don't know if it ever could be?? Then the whitelisting logic should be ignored and the Event Handlers should all be invoked. That makes the most sense to me.

PR #134, opened 2018-08-09
NIFIREG-192: Implement REGISTRY_START event

jdye64:


pvillard31:
Same here - code LGTM, will give it a try today or tomorrow if no one has the cycles to do it before.

nifi-registry-web-api/src/main/java/org/apache/nifi/registry/NiFiRegistryApiApplication.java
```diff
@@ -54,6 +63,15 @@ protected SpringApplicationBuilder configure(SpringApplicationBuilder applicatio
                 .properties(defaultProperties);
     }
 
+    @PostConstruct
+    public void postConstruct() {
```
kevdoran: I'm not sure when post construct gets called for a SpringApplicationBuilder. I don't think we want to fire the REGISTRY_START event until after the full application context is loaded, otherwise EventHookProvider consumers of that event type might not be loaded/ready yet.

Take a look at the `ApplicationListener<ApplicationReadyEvent>` interface, that could be a nice solution to this.

nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/NiFiRegistryResourceConfig.java
```diff
@@ -44,6 +48,14 @@
 
     private static final Logger logger = LoggerFactory.getLogger(NiFiRegistryResourceConfig.class);
 
+    @Autowired
+    private StandardProviderFactory standardProviderFactory;
+
+    @Bean
+    public EventService eventService() {
```
kevdoran: I don't think this is required because `EventService` is already annotated with `@Service`, so it will be created as a bean by component scanning.

kevdoran:
@jdye64 Thanks for that change. Now that #133 is merged, could you updated the documentation to add your new `REGISTRY_START` event type to the _Whitelisted Event Type x_ property value table?

jdye64:
Sure will. I will update that right now and make another commit

kevdoran:
+1, looks good and verified event is triggered when expected. Thanks for the contribution @jdye64!

PR #136, opened 2018-08-14
NIFIREG-194: Updated VersionedConnection to support load balancing co…

markap14:
…nfiguration

### DIFF
--- a/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/flow/VersionedConnection.java
+++ b/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/flow/VersionedConnection.java
@@ -35,6 +35,11 @@ public class VersionedConnection extends VersionedComponent {
     private List<String> prioritizers;
     private List<Position> bends;
 
+    private String loadBalanceStrategy;
+    private String partitioningAttribute;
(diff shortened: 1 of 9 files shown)


nifi-registry-data-model/src/main/java/org/apache/nifi/registry/flow/VersionedConnection.java
```diff
@@ -134,6 +139,36 @@ public void setPrioritizers(List<String> prioritizers) {
...
+            "unset or is set to any other value, the Partitioning Attribute has no effect.")
+    public String getPartitioningAttribute() {
+        return partitioningAttribute;
+    }
+
+    public void setPartitioningAttribute(final String partitioningAttribute) {
+        this.partitioningAttribute = partitioningAttribute;
+    }
+
+    @ApiModelProperty("Whether or not compression should be used when transferring FlowFiles between nodes")
```
bbende: Is there a list of allowable values for the types of compression?

markap14:
@bbende good catch. Yes, there should be allowable values added there. Just pushed another commit to address.

bbende:
Looks good to me, will merge, thanks!

PR #139, opened 2018-09-07
NIFIREG-198 Fix VersionedRemoteProcessGroup targetUri bug

kevdoran:
VersionedRemoteProcessGroup has two fields: targetUri and targetUris. They do not have simple getters, but rather each will fallback to using the other. Given how they are described, for certain combinations of targetUri and targetUris, they do not work correctly (mainly, when targetUri is not set and targetUris contains only a single value).

Furthermore, in NiFi only TargetUris is used/needed. It was a replacement for targetUri when the requirement for multiple URIs was added. Therefore, in addition to fixing the bug descirbed above, targetUri can be deprecated and dropped in future versions of NiFi Registry.


bbende:
+1 looks good and will merge, thanks!

PR #142, opened 2018-09-21
NIFIREG-200 Update dependencies

kevdoran:
- Update Jetty to version 9.4.11.v20180605
- Update Spring Boot to version 2.0.4.RELEASE
- Update Spring Security to version 5.0.7.RELEASE
- Update Jackson to version 2.9.6
- Drop Guava (was unused) from nifi-registry-web-api

### DIFF
--- a/nifi-registry-web-api/pom.xml
+++ b/nifi-registry-web-api/pom.xml
@@ -312,11 +312,6 @@
             <artifactId>jjwt</artifactId>
             <version>0.7.0</version>
         </dependency>
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>18.0</version>
-        </dependency>
         <!-- Test Dependencies -->
         <dependency>
             <groupId>org.apache.nifi.registry</groupId>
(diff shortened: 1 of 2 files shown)


kevdoran:
@bbende if you could review this when you get a chance that would be appreciated

alopresto:
@kevdoran any reason you decided on 18.0 for guava? Looks like the most recent release is 26.0 [Guava](https://github.com/google/guava).

kevdoran:
Good catch @alopresto. I was simply trying to pin a single version of Guava as we were using 18.0 but 17.0 was being pulled in transitively. But I don't see any reason not to pin version 26.0, so I updated my branch to use that. Compilation and runtime seem unaffected. Thanks!

bbende:
Reviewing..

bbende:
This looks good and I'll merge to master.

Something I noticed while reviewing this (not a result of anything in this PR), is that the nifi-registry-extensions modules are impacted by the top-level dependency management, which could have unforeseen consequences.

I created this JIRA with further information: https://issues.apache.org/jira/browse/NIFIREG-201

PR #144, opened 2018-10-30
NIFIREG-209 Rebuild metadata DB from FlowPersistenceProvider when emp…

bbende:
…ty DB AND when instance of MetadataAwareFlowPersistenceProvider

### DIFF
--- a/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/provider/flow/FlowMetadataSynchronizer.java
+++ b/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/provider/flow/FlowMetadataSynchronizer.java
@@ -1,0 +1,138 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.registry.provider.flow;
+
+import org.apache.nifi.registry.db.entity.BucketEntity;
+import org.apache.nifi.registry.db.entity.BucketItemEntityType;
+import org.apache.nifi.registry.db.entity.FlowEntity;
+import org.apache.nifi.registry.db.entity.FlowSnapshotEntity;
+import org.apache.nifi.registry.flow.FlowPersistenceProvider;
+import org.apache.nifi.registry.flow.MetadataAwareFlowPersistenceProvider;
+import org.apache.nifi.registry.metadata.BucketMetadata;
+import org.apache.nifi.registry.metadata.FlowMetadata;
+import org.apache.nifi.registry.metadata.FlowSnapshotMetadata;
+import org.apache.nifi.registry.service.MetadataService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.event.ContextRefreshedEvent;
+import org.springframework.context.event.EventListener;
+import org.springframework.stereotype.Component;
+
+import java.util.Date;
+import java.util.List;
+
+@Component
+public class FlowMetadataSynchronizer {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(FlowMetadataSynchronizer.class);
+
+    private MetadataService metadataService;
+    private FlowPersistenceProvider persistenceProvider;
+
+    @Autowired
+    public FlowMetadataSynchronizer(final MetadataService metadataService,
+                                    final FlowPersistenceProvider persistenceProvider) {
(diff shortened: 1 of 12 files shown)


bbende:
@kevdoran @ijokarumawak what do you guys think of this?

kevdoran:
This is a fun new feature @bbende. It will be nice for DR or as a mechanism to populate NiFi Registry instances in new environments with some canned flows. I looked at the code and tried it out and it's working well for me. There is some oddness when loading flows from 0.2.0 or 0.3.0 as they don't have all the metadata available, but I think that's fine. Going forward with git repos created after this feature exists should have everything there.

Interested to hear from @ijokarumawak on this to get his input as he is the most experienced with the Java Git stuff, but consider me a tentative +1 on this PR. Cool stuff!

ijokarumawak:
Oh, this is going to be very useful! I'm reviewing now. Thanks!

ijokarumawak:
@bbende This is really cool! I was able to:
1. Setup two NiFi Registry environment using the same remote Git repository, but different branches, dev and prod. 
2. Create new flow versions at dev and push it to remote origin/dev branch
3. Merge commits to origin/prod branch. Then remove database dir at prod environment to generate it from updated Git commits.
4. After restarting the prod NiFi Registry, I can import the new flows from dev to prod environment.
5. Updating existing flow scenario also worked. Even if I restored database from Git commits, NiFi was able to keep notice new version and change the deployed PG to a new version.

Haven't looked at the detail of implementation, but from functional point of view, I'm a huge +1! Great job!

MAliNaqvi:
@ijokarumawak 
how did you point the nifi registry to different branches? 
I am using docker-compose:

```
  nifi-registry:
    build:
      context: ./nifi-registry
    image: apache/nifi-registry
    ports:
      - 18080:18080
    volumes:
      - ./nifi-workflows:/opt/nifi-registry/nifi-workflows
    environment:
      - NIFI_REGISTRY_FLOW_STORAGE_DIR=/opt/nifi-registry/nifi-workflows
      - NIFI_REGISTRY_FLOW_PROVIDER=git
      - NIFI_REGISTRY_GIT_REMOTE=origin/DA-21-abcd
    env_file:
      - .env
```
The .env file contains the authN credentials.

```
NIFI_REGISTRY_GIT_USER
NIFI_REGISTRY_GIT_PASSWORD
```

but I get the error `Available remotes are [origin]`

kevdoran:
@MAliNaqvi that is not a supported feature. only one branch (typically master) can be used for tracking

ijokarumawak:
@MAliNaqvi I preconfigured the git directory manually before using it from NiFi registry. I used standard git commands. e.g. `git checkout -b dev` to create `dev` branch. I used two separate NiFi Registry instances for each `dev` and `prod` branch.

MAliNaqvi:
@ijokarumawak @kevdoran @bbende I applied the patch for this PR and for #152 and was able to get the nifi git sync working. I am able to create different branches eg. for separate JIRA issues and the registry pushes correctly and also creates the remote branch as well. This works really well.

Now, from what I understand, if we want to pull changes from git into a registry that is already running, we have to restart the registry which then repopulates its internal db. Is there another way? It would be great if the resync could be done via the nifi registry ui (ex. a button under the "Actions" drop down menu)

MAliNaqvi:
Also I am finding that restoring from git, then creating a commit, changes all group ids and makes numerous changes in the snapshot which prevents a clean diff of changes when creating PRs.

bbende:
@MAliNaqvi the use-case for this feature wasn't really to keep two registry instances in-sync by using git in between. It was more for the case where someone lost their registry instance, but still had the git repo, for example maybe running registry in a container with an H2 DB and pushing to github, and then the container goes away, so all that remains is the github repo.

As far as the diff'ing, the expectation is that diffs would be done through registry and not an external tool like github. There is already a REST end-point to compute the diff between two snapshots, but unfortunately the UI doesn't have anything yet to visualize this.

MAliNaqvi:
@bbende Thanks, I was able to use the REST endpoint to compute the diff.

bbende:
Seems like a few people have tested this out and signed off, so I'm going to merge this to master.

asxtray:
Has anyone tried the functionality from master branch? Pulled and rebuilt from master, staring registry with preconfigured GitFlowPersistenceProvider and local flow file git folder but the registry is empty. Shall I specify some startup flag to rebuild the database?

bbende:
@asxtray I haven't tried this from master in a while, but I don't know of any reason why it wouldn't still be working. Can you check your nifi-registry-app.log during start up for any of the following log statements?
```
"*  FlowPersistenceProvider is not metadata-aware, nothing to synchronize"
"*  Found existing buckets, will not synchronize metadata"
"*  Synchronizing metadata from FlowPersistenceProvider to metadata database..."
```
Let us know which of these printed.

asxtray:
It seems that flow folder(s) (bucket) must be the only top level folders in git. Now it grabs the bucket but skips the flows. I had to debug GitFlowMetaData to discover the problem. Lines 317, 318:
`final File flowSnapshotFile = new File(new File(backetFilePath).getParent(), flowSnapshotFilename);
final ObjectId objectId = flowSnapshotObjectIds.get(flowSnapshotFile.getPath());`
It relies on java.io.File object to construct flowSnapshot file git key, but the problem is that java.io.File is platform specific and delegates path construction to the internal file system abstraction. So for Linux it would be bucketName/flowfile.snapshot which is okay but for Windows it would be bucketName\flowfile.snapshot and it doesn't match the git key as it's unix style. What I did:
1. Provided bucketDirName as String backetFilePath into loadFlows
2. Replaced java.io.File manipulations with simple String concatenation:
`final String flowSnapshotFileGitKey = bucketDirName != null ? bucketDirName.concat("/").concat(flowSnapshotFilename) : flowSnapshotFilename;
final ObjectId objectId = flowSnapshotObjectIds.get(flowSnapshotFileGitKey);`

And that helped. Currently I don't have much time to submit a PR, but if I do I will, no guarantees.

PR #148, opened 2018-11-20
NIFIREG-211 Initial work for adding extenion bundles to NiFi Registry

bbende:
This PR includes the foundational work for adding extension bundles as a new type of versioned item. With this PR it is possible to upload NARs via the REST API with them being stored in the local filesystem persistence provider, and then interact with them through various REST APIs.

In the current state of this PR, registry does not have a way to know anything about the actual extensions contained in a bundle. This will need to come from a component descriptor provided within the NAR which will then be parsed and extracted on registry side.

We can leave NIFIREG-211 open after merging this so that we can continue to work against it and build out the additional functionality.

A NAR can be uploaded by making a multi-part form POST to the end-point:

_buckets/{bucketId}/extensions/bundles/{bundleType}_

An example using curl would be:

`curl -v -F file=@./nifi-example-processors-nar-1.0.0.nar http://localhost:18080/nifi-registry-api/buckets/65d6d70d-ecb7-478f-bbc4-1e378e52827c/extensions/bundles/nifi-nar`

From there you can interact with the REST API from the swagger docs.

I've also started a branch of NiFi with CLI commands related to extension bundles:
https://github.com/bbende/nifi/tree/extensions

If you run a full build of this PR first with mvn clean install, and then a build of the NiFi branch, you can launch the CLI in nifi-toolkit/nifi-toolkit-assembly/target/nifi-toolkit-1.9.0-SNAPSHOT-bin/nifi-toolkit-1.9.0-SNAPSHOT/bin/cli.sh and run the "upload-nars" command:

`registry upload-nars -b 024dba45-95c2-4a54-b79d-0abe1bb1122a -ebd /path/to/nifi-releases/nifi-1.8.0/lib -u http://localhost:18080 -verbose`

This should upload all of the NARs from the 1.8.0 release to your local registry.

Here is the commit history for all the changes made in this PR:

- Setting up DB tables and entities for extensions
- Updated MetadataService and DatabaseMetadataService with new methods for extension entities
- Added ON DELETE CASCADE to existing tables and simplified delete logic for buckets and flows
- Created data model for extension bundles and mapping to/from DB entities
- Created ExtensionBundleExtractor with an implemenetation for NARs
- Setup LinkService and LinkBuilder for extension bundles
- Setup pluggable persistence provider for extension bundles and implemented a local file-system provider.
- Refactored LinkService and add links for all extension related items
- Changed extension service to write out bundles to a temp directory before extracting and passing to persistence provider
- Implemented multi-part form upload for extensions bundles
- Upgraded to spring-boot 2.1.0
- Added SHA-256 checksums for bundle versions
- Initial client methods for uploading and retrieving bundles
- Configuring NiFi Registry Jersey client to use chunked entity processing so we don't load the entire bundle content into memory during an upload


bbende:
All versioned items (flows and now bundles) live in a bucket, and a bucket is where security policies are applied. So each bucket can be kind of like a mini extension repo, there could be a bucket for project nars and another bucket for other nars. 

The main rules are the following...

- Within a bucket, the bundle coordinate is unique, so you can't upload a NAR with the same group+artifact+version to the same bucket twice

- Across buckets you CAN upload NARs with the same group+artifact+version, BUT they must have the same checksum (currently SHA-256) which ensures they are actually the same bundle.

ottobackwards:
I don't want to pepper you with questions on this, is there some documentation that goes along with this?

bbende:
Before releasing any of this we would definitely have sections in the user guide and admin guide related to extension bundles, but we can't really write all that until the work is done, and this is only a starting point for the work.

Most of this is just integrating a new type of versioned item into the existing registry framework, so if you aren't already familiar with how registry works, then a starting point would probably be to play around with version controlling flows, and go through all the existing documentation, and then think of extension bundles as just another versioned thing like flows.

You can access the swagger documentation for the REST API from the running application:

http://localhost:18080/nifi-registry-api/swagger/ui.html

You can also get a feel for how the current API works from looking at this integration test:

https://github.com/apache/nifi-registry/blob/7744072b5f059fe5240746e968493eddefa719a6/nifi-registry-core/nifi-registry-web-api/src/test/java/org/apache/nifi/registry/web/api/UnsecuredNiFiRegistryClientIT.java#L320-L469

nifi-registry-core/nifi-registry-framework/src/main/resources/db/migration/V3__AddExtensions.sql
```diff
@@ -0,0 +1,62 @@
...
+    CONSTRAINT FK__EXTENSION_BUNDLE_BUCKET_ITEM_ID FOREIGN KEY (ID) REFERENCES BUCKET_ITEM(ID) ON DELETE CASCADE,
+    CONSTRAINT FK__EXTENSION_BUNDLE_BUCKET_ID FOREIGN KEY(BUCKET_ID) REFERENCES BUCKET(ID) ON DELETE CASCADE,
+    CONSTRAINT UNIQUE__EXTENSION_BUNDLE_BUCKET_GROUP_ARTIFACT UNIQUE (BUCKET_ID, GROUP_ID, ARTIFACT_ID)
+);
+
+CREATE TABLE EXTENSION_BUNDLE_VERSION (
+    ID VARCHAR(50) NOT NULL,
+    EXTENSION_BUNDLE_ID VARCHAR(50) NOT NULL,
+    VERSION VARCHAR(100) NOT NULL,
+    DEPENDENCY_GROUP_ID VARCHAR(500),
```
kevdoran: CPP extensions may have more than one dependency, in which case having these in a separate table may be worthwhile.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/ExtensionRepositoryResource.java
```diff
@@ -0,0 +1,378 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(
+            value = "Gets the names of the buckets the current user is authorized for in order to browse the repo by bucket",
+            response = ExtensionRepoBucket.class,
+            responseContainer = "List",
+            extensions = {
+                    @Extension(name = "access-policy", properties = {
+                            @ExtensionProperty(name = "action", value = "read"),
+                            @ExtensionProperty(name = "resource", value = "/buckets/{bucketId}") })
+            }
```
kevdoran: I don't think we want an access-policy documented here as read only access is allowed for all authenticated users.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/extension/StandardExtensionService.java
```diff
@@ -0,0 +1,589 @@
...
+    }
+
+    private ExtensionBundleEntity getOrCreateExtensionBundle(final String bucketId, final String groupId,
+                                                             final String artifactId, final ExtensionBundleType bundleType) {
+        ExtensionBundleEntity existingBundleEntity = metadataService.getExtensionBundle(bucketId, groupId, artifactId);
+        if (existingBundleEntity == null) {
+            final ExtensionBundle bundle = new ExtensionBundle();
+            bundle.setIdentifier(UUID.randomUUID().toString());
+            bundle.setBucketIdentifier(bucketId);
+            bundle.setName(groupId + " - " + artifactId);
```
kevdoran: Don't know how others feel about this, but personally I would have a slight preference for using the Maven coordinate style here for the inferred bundle name, i.e.:

    bundle.setName(groupId + ":" + artifactId);

nifi-registry-core/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/extension/ExtensionBundleVersionMetadata.java
```diff
@@ -0,0 +1,161 @@
...
+    public long getTimestamp() {
+        return timestamp;
+    }
+
+    public void setTimestamp(long timestamp) {
+        this.timestamp = timestamp;
+    }
+
+    @ApiModelProperty(value = "The identity that created this version")
+    public String getAuthor() {
```
alopresto: Understanding that "authors" can take many forms, is this a standardized String representation, or could it be "alopresto", "Andy LoPresto", and "Andy LoPresto <alopresto@apache.org>" are all valid and used simultaneously, causing lookup/resolution issues later? Would using a custom datatype here make sense?

nifi-registry-core/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/extension/ExtensionBundleVersionMetadata.java
```diff
@@ -0,0 +1,161 @@
...
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+
+        final ExtensionBundleVersionMetadata other = (ExtensionBundleVersionMetadata) obj;
+        return Objects.equals(this.id, other.id);
```
alopresto: Will this cause issues with `Set`/`List` assignments/`#contains()` because the IDs can be equal but the data within them different? I remember this was always an issue with Java O/RM.

nifi-registry-core/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/extension/ExtensionBundleVersionMetadata.java
```diff
@@ -0,0 +1,161 @@
...
+    @Min(1)
+    private long timestamp;
+
+    @NotBlank
+    private String author;
+
+    private String description;
+
+    @NotBlank
+    private String sha256Hex;
```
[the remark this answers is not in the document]
alopresto: How easy is it to add a new field here? I'd like to propose storage of a cryptographic signature of some sort, in addition to the SHA-256 checksum, which can be modified (generated over new contents) by a malicious user. What inputs (data, fields, encoding, etc.) does the SHA-256 cover?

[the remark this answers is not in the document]
bbende: @kevdoran I'm good with calling it sha256 for consistency

@alopresto It is fairly straight forward to add another field here, just typical changes like DB schema update, new member field, modify mapping code (no ORM here :( ). Right now the input of the SHA-256 is the binary content of the bundle (i.e. the input stream of actual NAR file). 

Did you want me to add in the cryptographic signature in this PR (assuming you point at how to calculate it), or do you want to wait til this is merged and add as follow on work? 

It can be done either way, there is still a bit more work to do before this is ready anyway.

nifi-registry-core/nifi-registry-framework/src/main/resources/db/migration/V3__AddExtensions.sql
bbende: That is a good point, it probably makes sense to separate out a separate table where bundle version can have one-to-many dependencies, and NARs will just so happen to only ever have one. Will work on this update.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/extension/StandardExtensionService.java
bbende: Makes sense, I'm on-board with that

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/ExtensionRepositoryResource.java
bbende: That sounds consistent with the getBuckets end-point, so I'll remove that.

nifi-registry-core/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/extension/ExtensionBundleVersionMetadata.java
bbende: Also, unfortunately we have no ORM here and are writing our own JDBC :(

nifi-registry-core/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/extension/ExtensionBundleVersionMetadata.java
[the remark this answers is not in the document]
alopresto: I would like to have a conversation around the cryptographic signatures so if it's possible to add a new field fairly easily, I'm ok waiting for this to be in. I am curious if people think there should be any checksum coverage on the metadata for the extension as well. I like having a calculation over the binary, but I also think there is value in ensuring the metadata values are checked as well. Looking for feedback, but this should be independent from this PR.

nifi-registry-core/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/extension/ExtensionBundleVersionMetadata.java
alopresto: Understood. I think that long-term, we probably need to re-evaluate how that is being done, as the identities between NiFi and NiFi Registry may not always match.

nifi-registry-core/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/extension/ExtensionBundleVersionMetadata.java
alopresto: Works as expected.

nifi-registry-core/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/extension/ExtensionBundleVersionMetadata.java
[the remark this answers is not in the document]
alopresto: Ok. Is there any way to craft an instance of a binary with different metadata that generates the same checksum? Can you point me to the code which performs this inspection and calculation? Where does the checksum reside in transmission (i.e. the binary and checksum are sent separately, or it's just calculated on one end and there is nothing to compare it to?)?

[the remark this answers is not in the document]
bbende: Right now it is only calculated on the NiFi Registry side and would be used by consumers, such as NiFi, to verify they have downloaded the correct bundle:

https://github.com/bbende/nifi-registry/blob/extensions/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/extension/StandardExtensionService.java#L146-L153

The InputStream there is from a REST end-point where you make a multipart form POST, which is here:

https://github.com/bbende/nifi-registry/blob/extensions/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketExtensionResource.java#L107-L125

Do you think we should be requiring the SHA-256 to be submitted along with the file? This way we can verify that what we received was what was expected to be sent? 

We don't have anything like that when saving flows so I hadn't thought of it that way.

bbende:
@kevdoran just pushed a commit that addresses the review comments, let me know if you find anything else. As an FYI, I did try to change that swagger annotation for LinkableEntity, but it doesn't seem to like Link.JaxbLink or the fully qualified classname, so will leave that for another time.

The only outstanding question from the conversation yesterday was whether the current SHA-256 approach is acceptable, or if we clients she be providing that when uploading a bundle. Need @alopresto 's thoughts on that.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/extension/StandardExtensionService.java
```diff
@@ -188,13 +188,14 @@ public ExtensionBundleVersion createExtensionBundleVersion(final String bucketId
...
             final String userIdentity = NiFiUserUtils.getNiFiUserIdentity();
+            final long bundleCreatedTime = extensionBundle.getCreated().getTime();
 
             final ExtensionBundleVersionMetadata versionMetadata = new ExtensionBundleVersionMetadata();
             versionMetadata.setId(UUID.randomUUID().toString());
             versionMetadata.setExtensionBundleId(extensionBundle.getId());
             versionMetadata.setBucketId(bucketIdentifier);
             versionMetadata.setVersion(version);
-            versionMetadata.setTimestamp(System.currentTimeMillis());
+            versionMetadata.setTimestamp(bundleCreatedTime);
```
kevdoran: This works great when the bundle is being created with the first version uploaded. But I noticed when subsequent versions are uploaded they are also getting tagged with the same creation time. Given that we don't know what was done in getOrCreateExtensionBundle, we might need to refactor the code above, or... it might be better to just revert this change to always use currentTime for the bundle version. I'm good with either

bbende:
Just pushed a new commit that allows the client to specify a SHA-256 when uploading the bundle. If they specify one and it doesn't match what the server calculates, then it is rejected. If they don't specify one then we accept the bundle, but we mark a flag called "sha256Supplied" as false, so consumers later will know if the bundle was uploaded with or without a client provided checksum.

Since the upload is already a multi-part request, I thought it made the most sense for the SHA-256 to be another field in the multi-part form data. As an example of a curl command to upload with the checksum:

`curl -v -F file=@./nifi-example-processors-nar-1.0.0.nar -Fsha256=93d2c6537142497ca3f908452d3c3c2bcae928fdc07e1b48fd3ba95ec22bc639 "http://localhost:18080/nifi-registry-api/buckets/72f897d9-5645-4168-852f-4885cc7c9cfe/extensions/bundles/nifi-nar"
`

PR #149, opened 2018-12-11
NIFIREG-215 Extension Bundle Improvements

bbende:
I left the changes as separate commits in order to more easily review each change on it's own, but feel free to squash upon merging.


kevdoran:
Reviewing...

nifi-registry-core/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/bucket/Bucket.java
```diff
@@ -41,6 +41,8 @@
 
     private String description;
 
+    private Boolean allowExtensionBundleRedeploy;
```
kevdoran: Instead of a boolean flag here. I'm wondering if this would be better as a Enum for something like "redeploy policy" with values such as "NEVER, SNAPSHOTS, ALWAYS" or something like that?

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/ExtensionResource.java
```diff
@@ -93,15 +94,25 @@ public ExtensionResource(final RegistryService registryService,
...
-    public Response getExtensionBundles() {
+    public Response getExtensionBundles(
+            @QueryParam("groupId")
+            @ApiParam("Optional groupId to filter results. The value may be an exact match, or a trailing wildcard, " +
+                    "such as 'com.%' to select all bundles where the groupId starts with 'com.'.")
+                final String groupId,
+            @QueryParam("artifactId")
+            @ApiParam("Optional artifactId to filter results. The value may be an exact match, or a trailing wildcard, " +
+                    "such as 'nifi-%' to select all bundles where the artifactId starts with 'nifi-'.")
+                final String artifactId) {
```
kevdoran: These parameters are nice to have, and work as expected for me. One thing I noticed is that the documentation says _"... a trailing wildcard ..."_, which indicates wildcards can only be used at the end of the value. However in testing, I was able to use them anywhere in the value (at the beginning or middle of the value) and it worked (with the expected results as well).

So maybe just update the documentation to remove the work "trailing"?

nifi-registry-core/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/bucket/Bucket.java
kevdoran: Ok, I’m not familiar with Nexus’s configuration options, but I’m in favor of keeping consistencies/parallels with that. I’m good with leaving this as-is.

bbende:
@kevdoran I pushed up a commit to update the filter params documentation. I decided not to do anything with the lastModified value on the bundle table because technically nothing on that row is ever changing, it is really the entries in the version table which are changing, and this will have new create dates when redeploying any version. Let me know if there is anything else.

kevdoran:
Thanks, @bbende! +1, merged to master.

PR #150, opened 2018-12-12
Fixes bug in path to providers file

joaocc:


kevdoran:
Thanks for the contribution @joaocc! There was another PR with the same set of changes. Unfortunately can only merge one, so I closed this PR and added you as a co-author on the other commit with the same changes.

PR #151, opened 2018-12-12
Bugfix overriding db props via environment variables (Docker)

jgondron:
The mapping from env to registry properties for the database user, pass, and max connections are referencing the incorrect property names. This should fix the issue in the Docker update_database.sh script, and correct other references in the documentation.


kevdoran:
+1, merging. Sorry for the delay in getting around to testing this, been busy focusing on other efforts.

PR #152, opened 2018-12-14
Bugfix overriding flow provider via environment variables (Docker)

jgondron:
The script which maps env to flow provider properties is currently not overwriting the file, but instead sending to stdout. This should fix the issue in the Docker update_flow_provider script.

Note: This replicates a fix also made in https://github.com/apache/nifi-registry/pull/132


kevdoran:
Thanks @jgondron! Tested it and merged to master

PR #153, opened 2018-12-18
NIFIREG-220 New test certs

kevdoran:
Replacing the test TLS certs in the source code repository
with new certs that were signed by a tracked CA key. Also
adding instructions for generating additional keys in the
future if needed.


alopresto:
Reviewing...

nifi-registry-core/nifi-registry-web-api/src/test/resources/application-ITSecureFile.properties
```diff
@@ -28,9 +28,9 @@ nifi.registry.client.properties.file: src/test/resources/conf/secure-file/nifi-r
 
 # Embedded Server SSL Context Config
 server.ssl.client-auth: need
-server.ssl.key-store: ./target/test-classes/keys/localhost-ks.jks
-server.ssl.key-store-password: localhostKeystorePassword
-server.ssl.key-password: localhostKeystorePassword
+server.ssl.key-store: ./target/test-classes/keys/registry-ks.jks
```
alopresto: Is there a reason for the inconsistent formatting of these keys? Ideally, I'd prefer if `ssl` was changed to `tls`, but I understand legacy naming. However, the `.` and `-` seem odd.

nifi-registry-core/nifi-registry-web-api/src/test/resources/application-ITSecureFile.properties
```diff
@@ -28,9 +28,9 @@ nifi.registry.client.properties.file: src/test/resources/conf/secure-file/nifi-r
...
+server.ssl.key-store-password: password
```
alopresto: These passwords seem to have gotten much weaker.

nifi-registry-core/nifi-registry-web-api/src/test/resources/keys/README.md
```diff
@@ -12,36 +12,217 @@
...
+| CN=user1, OU=nifi | user1-ks.p12 | client (user="user1") key/cert keystore | PKCS12 | password |
 
-    keys/
-     +-- client-ks.jks      # client keystore: keystorePass=clientKeystorePassword, keyPass=u1Pass
-     +-- localhost-ks.jks   # server keystore: keystorePass=localhostKeystorePassword, keyPass=localhostKeystorePassword
-     +-- localhost-ts.jks   # server/client truststore (contains CA): truststorePass=localhostTruststorePassword
+## Generating Additional Test Keys/Certs
 
-Copy these files to the test/resources/keys/ directory.
+If we need to add a service to our test environment that requires a cert signed by the same CA, here are the steps for generating additional keys for thsi directory that are signed by the same CA key.
```
alopresto: Typo - _thsi_ -> _this_

nifi-registry-core/nifi-registry-web-api/src/test/resources/keys/README.md
```diff
@@ -12,36 +12,217 @@
...
 
+Requirements:
+
+- docker
+- keytool (included with Java)
+- openssl (included/available on most platforms)
+
+If you do not have docker, you can alternatively use the nifi-toolkit binary which has a dependency on Java 1.8 and is available for download from http://nifi.apache.org.
```
alopresto: Use HTTPS protocol - _https://nifi.apache.org_

nifi-registry-core/nifi-registry-web-api/src/test/resources/keys/README.md
```diff
@@ -12,36 +12,217 @@
...
+
+If you do not have docker, you can alternatively use the nifi-toolkit binary which has a dependency on Java 1.8 and is available for download from http://nifi.apache.org. 
+
+### New Service Keys
+
+The steps for generating a new *service* key/cert pair are (using `proxy` as the example service):
+
+```
+# make working directory
+rm -rf /tmp/test-keys
```
alopresto: Is removing the directory necessary? We should include a note explaining what will happen here. While it's obvious to most users, some may have generated keys in that directory that they are still using, etc.

nifi-registry-core/nifi-registry-web-api/src/test/resources/keys/README.md
```diff
@@ -12,36 +12,217 @@
...
+      --hostnames proxy \
+      --subjectAlternativeNames localhost \
+      --nifiDnSuffix ", OU=nifi" \
+      --keyStorePassword password \
+      --trustStorePassword password \
+      --days 9999 \
+      -O
+
+# switch to output directory, create final output directory
+cd /tmp/test-keys
```
alopresto: Isn't the user still in `/tmp/test-keys` at this point?

nifi-registry-core/nifi-registry-web-api/src/test/resources/application-ITSecureFile.properties
kevdoran: These are spring boot application properties names that are getting interpreted by that framework:
https://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html

nifi-registry-core/nifi-registry-web-api/src/test/resources/keys/README.md
kevdoran: will update

nifi-registry-core/nifi-registry-web-api/src/test/resources/keys/README.md
kevdoran: will correct

nifi-registry-core/nifi-registry-web-api/src/test/resources/keys/README.md
kevdoran: yeah just had it there in case they did not run these command all together. it should not hurt if they are already there

nifi-registry-core/nifi-registry-web-api/src/test/resources/keys/README.md
kevdoran: Yeah I was on the fence about that. In the end I was more worried that they would be copying old keys to the final output directory mistakenly. could be convinced either way...

nifi-registry-core/nifi-registry-web-api/src/test/resources/application-ITSecureFile.properties
kevdoran: Agree on the unfortunate legacy naming used by all these libs/frameworks though. it's unfortunate 😞

nifi-registry-core/nifi-registry-web-api/src/test/resources/keys/README.md
alopresto: Understood. I guess it's ok because everything can be recreated, but I just worry about providing users with copy/paste commands that take destructive action in the event those users don't understand the consequences. Theoretically, one would never have to worry about that in a security context, setting up keystores and certificates, but obviously it happens. 

Could provide a script which named the subdirectory using something like `$(date +"%T")` so they are unique, but it might make it harder for users to know which subdirectory to copy from.

kevdoran: @alopresto the latest commit addresses this. Feel free to squash commits on merge and discard the fix commit messages.

alopresto:
Ran `contrib-check` and all tests (including _integration tests_) pass. +1, merging.

PR #157, opened 2019-02-25
NIFIREG-229: Handle error in LDAP sync background thread

kevdoran:
- Catch thrown errors and exceptions in background ldap sync task to prevent killing the thread
- Improve logging by moving some warns to debug for potentially valid configurations and add more debug output.


mcgilman:
Will review...

PR #159, opened 2019-03-01
NIFIREG-232 Introducing flow persistence provider migrator and toolkit

brosander:


bbende:
@brosander this looks interesting, will definitely take a closer look soon

brosander:
I probably need to add a readme section if this pr looks good to you @bbende 

Usage:
1. Shutdown registry
1. Copy providers.xml -> providers-to.xml
1. In registry home as working dir, run persistence-toolkit.sh -t providers-to.xml

I tested it by converting filesystem provider to git provider, then back to filesystem provider and diffing original, roundtrip directory.

Also by using it on a registry and verifying that everything looked right after conversion (could still get old versions of things, sync w/ nifi, etc)

nifi-registry-toolkit/nifi-registry-toolkit-persistence/src/main/java/org/apache/nifi/registry/toolkit/persistence/FlowPersistenceProviderMigrator.java
```diff
@@ -0,0 +1,106 @@
...
+
+                    log.info("Migrated flow {} version {}", flow.getName(), version);
+                }
+            }
+        }
+    }
+
+    public static void main(String[] args) {
+        Options options = new Options();
+        options.addOption("t", "to", true, "Providers xml to migrate to.");
```
bbende: Would it make sense to have an argument to specify the nifi-registry home dir so that the script didn't have to be executed from already being in that location?

I'm just thinking how my normal thought would be to navigate to where the toolkit is and then run something from bin.

I guess if we did change it allow specifying the location of nifi-registry, then the tool itself has to cd to that location first since config in the conf dir could be referencing relative paths.

brosander: The relative paths in registry conf was why I implemented it the way I did, we could do something like changing working directory before any logic though if you think it'd make it more useable.

nifi-registry-toolkit/nifi-registry-toolkit-assembly/NOTICE
```diff
@@ -0,0 +1,18 @@
...
+
+This includes derived works from the Apache NiFi (ASLv2 licensed) project (https://git-wip-us.apache.org/repos/asf?p=nifi.git):
+  Copyright 2015-2018 The Apache Software Foundation
+  This includes sources for bootstrapping, runtime, component API, security/authorization API
+
+This includes derived works from the Apache NiFi MiNiFi (ASLv2 licensed) project (https://git-wip-us.apache.org/repos/asf/nifi-minifi.git):
+  Copyright 2015-2018 The Apache Software Foundation
+  This includes build configuration for the Swagger UI.
+
+(ASLv2) Apache Commons CLI
```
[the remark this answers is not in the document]
brosander: Good point

nifi-registry-toolkit/nifi-registry-toolkit-persistence/src/main/java/org/apache/nifi/registry/toolkit/persistence/FlowPersistenceProviderMigrator.java
brosander: Looks like there's not a great way to do that from inside Java, we'd probably need to add special handling for that argument w/in the shell scripts to support it.

brosander:
I can add to the admin guide and update the doc around changing flow persistence providers as well.

nifi-registry-toolkit/nifi-registry-toolkit-persistence/src/main/java/org/apache/nifi/registry/toolkit/persistence/FlowPersistenceProviderMigrator.java
bbende: I think as long as we clearly show how to run it then we probably leave it as is.

bbende:
Thanks for addressing the feedback. For the documentation, we can probably just have a README in the nifi-registry-toolkit module, and then if we end up adding more tools and have a significant amount of content, we can promote it to a nifi registry toolkit guide like we did in NiFi.

brosander:
@bbende I think I've addressed your feedback, please let me know if more changes are needed

bbende:
@brosander looks good, thanks! Going to merge

PR #162, opened 2019-03-14
NIFIREG-238: Allow aliasing of registry url

brosander:


brosander:
@bbende @kevdoran mind taking a look when you get a chance?

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/alias/RegistryUrlAliasService.java
```diff
@@ -0,0 +1,167 @@
...
+                // find the schema
+                final SchemaFactory schemaFactory = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
+                final Schema schema = schemaFactory.newSchema(StandardProviderFactory.class.getResource(ALIASES_XSD));
+
+                // attempt to unmarshal
+                final Unmarshaller unmarshaller = JAXB_CONTEXT.createUnmarshaller();
+                unmarshaller.setSchema(schema);
+
+                // set the holder for later use
+                final JAXBElement<Aliases> element = unmarshaller.unmarshal(new StreamSource(configurationFile), Aliases.class);
```
alopresto: I believe this construct would be vulnerable to an XXE attack. Please see NiFi's [`AuthorizerFactoryBean.java`](https://github.com/apache/nifi/blob/master/nifi-nar-bundles/nifi-framework-bundle/nifi-framework/nifi-authorizer/src/main/java/org/apache/nifi/authorization/AuthorizerFactoryBean.java#L197) for an example of how to mitigate this.

alopresto:
I have some questions on this feature:

* How are aliases set? It appears that this only happens from an internal XML file at this time, but if the REST API is extended to handle this in the future, it exposes a number of security concerns. 
* What happens if the value inside changes without the end user knowing (i.e. Kevin versions two PGs, one called *Do normal stuff* and it references *Do sensitive stuff*, and then Andy modifies the alias in *Do normal stuff* that points to *Do sensitive stuff* to point to _Andy’s version_ of *Do sensitive stuff* and Kevin doesn’t know about it)?
* I understand the example `aliases.xml` is demonstrating an _internal_ token and an _external_ URL, but I think using `http://localhost:18080` for the URL is confusing as for most common understandings of those terms, it would be "internal", not "external". Perhaps a value like `http://registry.nifi.apache.org:18080` would illustrate this point more clearly.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/alias/RegistryUrlAliasService.java
brosander: Will apply mitigation

brosander:
Thanks for the feedback @alopresto 

1. I'd intended it to be configured only by OS users able to administer the registry, similarly to flow persistence providers for example.
2. I'd expect it to be the same as if the admin edited the flowfile content directly on disk (since they have access to the registry install).  I'm not sure how NiFi behaves if the registry starts returning a modified flow without a version change.
3. I'll update the example as suggested.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/alias/RegistryUrlAliasService.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private final List<Pair<String, String>> aliases;
+
+    @Autowired
+    public RegistryUrlAliasService(NiFiRegistryProperties niFiRegistryProperties) {
+        this(createAliases(niFiRegistryProperties));
+    }
+
+    private static List<Pair<String, String>> createAliases(NiFiRegistryProperties niFiRegistryProperties) {
```
alopresto: Is there a reason this returns `List<Pair<String, String>>` instead of a `Map<String, String>`? If sorting is a concern, `SortedMap<>` is available.

brosander: It makes migration from URLs in the registry coordinates to a token possible while still allowing nifi to point at older versions of flows.

Ex:
```
<aliases>
    <alias>
        <internal>INTERNAL_NIFI_REGISTRY_TOKEN</internal>
        <external>https://my.new.secure.registry.url:18443</external>
    </alias>
    <alias>
        <internal>http://my.old.insecure.registry.url:18080</internal>
        <external>https://my.new.secure.registry.url:18443</external>
    </alias>
</aliases>
```

Should allow you to spin up the registry, nifi pointing at the new url able to point at arbitrary old revisions of flows.  If you write them back they should go into the registry with INTERNAL_NIFI_REGISTRY_TOKEN.

alopresto: Does either the internal or external value have a unique constraint enforced? If the internal collection does not have a unique constraint enforced, what happens if there is a value collision there?

brosander: I think a unique constraint on the internal token is a good idea, will add.

brosander:
@alopresto I believe I've addressed your feedback, please let me know if further changes are needed or if there's anything else I can clarify.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/service/alias/RegistryUrlAliasService.java
alopresto: I'm still seeing this as `List<Pair<String, String>>` -- did I miss a commit?

brosander: @alopresto Changed to a LinkedHashMap

brosander:
@alopresto please let me know if there are additional changes needed.

alopresto:
I believe the security concerns I brought up have been addressed, provided the method of modification is still _OS user with write access to these internal representations_. If this functionality becomes exposed over the REST API, I have additional concerns.

bbende:
Taking a look as well...

nifi-registry-core/nifi-registry-properties/src/main/java/org/apache/nifi/registry/properties/NiFiRegistryProperties.java
```diff
@@ -61,6 +61,7 @@
     public static final String EXTENSION_DIR_PREFIX = "nifi.registry.extension.dir.";
 
     public static final String PROVIDERS_CONFIGURATION_FILE = "nifi.registry.providers.configuration.file";
+    public static final String REGISTRY_ALIAS_CONFIGURATION_FILE = "nifi.registry.registry.alias.configuration.file";
```
bbende: We also need to update the default properties file that comes in the assembly to have this property

brosander:
@kevdoran Thanks for the feedback, here are my thoughts around the suggestion.

TL;DR - I'd prefer to keep this change as-is and evaluate all xml config files as part of a larger usability effort around all the JAX-B config but will make the change if everyone feels that we should add it to properties instead.

Rationale:

I was concerned with adding a ton of extra properties to the nifi-registry.properties file and decided to follow convention for how other things seemed to be done.

It could be implemented with properties but for my particular usecase [order of definition matters](https://github.com/apache/nifi-registry/pull/162#discussion_r269328710) and the properties will come in unordered.

This would leave the user responsible for defining property names that sort alphabetically (or however we define the sorting logic) into the order they want their aliases in an upconvert case instead of a more natural list of properties.

To be user friendly, I'd think we'd have to have actual numbers instead of names determine the order:
```
nifi.registry.alias.1.internal = INTERNAL_NIFI_REGISTRY_TOKEN
nifi.registry.alias.1.external = https://my.new.secure.registry.url:18443 
nifi.registry.alias.2.internal = http://my.old.insecure.registry.url:18080
nifi.registry.alias.2.external = https://my.new.secure.registry.url:18443
```

If we still wanted to name them, we'd need either a convention property name (nifi.registry.alias.1.human_name.internal, nifi.registry.alias.1.human_name.external) or an additional property (nifi.registry.alias.1.name).

nifi-registry-core/nifi-registry-properties/src/main/java/org/apache/nifi/registry/properties/NiFiRegistryProperties.java
brosander: Think I've resolved with https://github.com/apache/nifi-registry/pull/162/commits/a2ab149394971b94a1fc571ea340268ed9bc1623

kevdoran:
@brosander That's fair, those are good reasons. I admit I had not considered order in this case.

FWIW, Identity Mappings are also order dependent and are sorted alpha-numerically based on the "key/name" portion of the property. I agree it is not intuitive in that case either (though it is also something that users have been exposed to since Identity Mappings were introduced in NiFi).

If/when we have a chance to do a larger configuration refactoring with breaking changes, I'd suggest we look closely at this. I've seen a lot of ways projects manage order-dependent properties, including: 

1. alpha-numeric sorting of the property keys that make up the collection
2. preserving order of properties set in the properties file (requires a custom Properties loader)
3. using array index format, such as:

```
nifi.registry.aliases[0].internal = INTERNAL_NIFI_REGISTRY_TOKEN
nifi.registry.aliases[0].external = https://my.new.secure.registry.url:18443 
nifi.registry.aliases[1].internal = http://my.old.insecure.registry.url:18080
nifi.registry.aliases[1].external = https://my.new.secure.registry.url:18443
```

bbende:
Sounds like we can keep this as is for now, and always do a larger re-evaluation of properties later. The latest changes look good, so I'll go ahead and get this merged. Thanks for implementing this and making the changes.

brosander:
Thanks for merging, appreciate the feedback everyone!

PR #164, opened 2019-04-01
NIFIREG-245 Refactor REST API docs

bbende:
This PR changes the way the REST API documentation is generated to leverage the swagger2markup Maven plugin which produces asciidoc from the swagger.json, and then we can convert the asciidoc to html the same way we do for our other docs.

In addition, this PR standardizes the use of all the @ ApiOperation annotations so that the main value is a short description closely resembling the Java method name, and the notes fields contains any further explanation.

NOTE: This PR is based off the PR for NIFIREG-233 since there were additional REST methods in that PR that would be impacted by this PR.


nifi-registry-core/nifi-registry-web-api/src/main/resources/templates/index.html.hbs
```diff
@@ -477,6 +481,38 @@
...
+            <div id="extension-endpoints" class="section-endpoints hidden"></div>
+        </div>
+        <div class="section">
+            <div class="section-header">
+                <div class="title link">Extension Repository</div>
+                <div class="sub-title section-description">Extension Repository endpoints</div>
+                <div class="clear"></div>
+            </div>
+            <div id="extension-repository-endpoints" class="section-endpoints hidden"></div>
+        </div>
```
kevdoran: Are we keeping the Handlebars-based REST API documentation in addition to the new ascii-doc based HTML, or can this go away entirely?

bbende: I figured it didn't hurt to leave the old approach around for at least one release in case we needed to switch back for some reason. With everything as is, all we would have to do is switch the link on the left side of the main documentation.jsp to point at the old index.html rather than the new rest-api.html.

kevdoran:
Rebasing my local clone now that #163 is merged and will merge this momentarily.

PR #165, opened 2019-04-01
NIFIREG-247 Adding ScheduleState enum and adding field to processor, …

bbende:
…port, and remote port


kevdoran:
Thanks, @bbende. Question: should this new field be added to Controller Service as well?

bbende:
@kevdoran thanks for taking a look... processors and ports are the only components that have three levels of states (i.e. disabled, stopped, running). Controller services are enabled or disabled which is basically started or stopped, reporting tasks are started or stopped, and RPGs are transmission enabled or disabled. So I think we only need to capture this to tell the difference between disabled and stopped for the components that support three states.

PR #167, opened 2019-04-02
NIFIREG-248 Fixing usage of the JAX-RS Link class so that it works co…

bbende:
…rrectly with swagger code-gen


kevdoran:
Thanks @bbende. Reviewing...

PR #169, opened 2019-04-09
NIFIREG-216 Add S3 BundlePersistenceProvider

bbende:
This PR also includes some refactoring of the BundlePersistenceProvider API so I left it as two commits in case it was easier to review how the changes evolved.

To test this you can build with -Pinclude-aws and follow the README in ext/aws/. Pre-requisites would be an AWS account with credentials configured in one of the standard ways and an existing bucket.


kevdoran:
Will review

nifi-registry-assembly/pom.xml
```diff
@@ -423,5 +423,43 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>include-aws</id>
```
kevdoran: I like making this provider an optional extension, but is there any reason not to include it by default (i.e., licensing or binary size).

Or do you think for our Apache convenience binary builds we will include this profile? I think it would be pretty popular. Just wondering if we need to discuss before the next release, and maybe update the RM guide to specify if this profile should be active for convenience binary builds.

bbende:
@kevdoran thanks for the review, I'm taking a look at your PR right now and will get it incorporated into this branch. Adding the URL option makes total sense and seems so obvious in hindsight.

As far as the assembly/profile... I would be fine making the include-aws profile enabled by default, or documenting to enable the profile during releases. I started down the path of how the Ranger assembly was setup, but once I was done and looked at it, I started wondering whether the entire setup was really necessary and whether all the AWS stuff should just go in nifi-registry-framework. The only dependencies that may eventually need to be isolated seem like the jackson dependencies, since it could be possible that the AWS client at some point requires a different version than registry core. So I figured maybe it is still good to keep the structure for possible isolation down the road.

bbende:
I merged in the PR for the URL and docker work so this branch is now updated.

I was playing around with the include-aws profile, and if we want to include it by default we can do something like:
```
<id>include-aws</id>
            <activation>
                <property>
                    <name>!skipAws</name>
                </property>
            </activation>
```
The only downside I see is that users still need to go into nifi-registry.properties and add `nifi.registry.extension.dir.aws=./ext/aws/lib`. We could have that in the properties file commented out to make it easier, or we could try to do something during the build to conditionally inject it into the properties file, or just leave it up to reading documentation.

kevdoran:
Hi @bbende, I like keeping it optional and structured as an extension. That gives us the most flexibility.

I don't have a strong opinion the build profile being enabled by default or not. If we do want it enabled by default, I like doing it the way you propose in your comment above, If it is not enabled by default, I do think we should update the release instructions to explicitly enable it. My reasoning is that  (1) I think it will be a popular feature, and (2) the docker image is usually built from the convenience binary, and I would like the S3 provider to be available in the `apache/nifi-registry` dockerhub image. 

Good point regarding the extension dir property. Is this an option: 

- nifi-registry.properties gets this added:

      nifi.registry.extension.dir.aws=${nifi.registry.extension.dir.aws}

- when the AWS profile is active, that property gets set to `./ext/aws/lib` in the pom
- when the AWS profile is not active, that property gets set to empty string in the pom
- the code gets updated (if it is not already the case) to allow / ignore blank values for extension dir properties (perhaps logging to info or warn during startup to make it clear that no value was set?)

bbende:
@kevdoran I pushed a new commit that includes the aws extensions by default using the approach you described. Let me know if there is anything else to update or change.

PR #178, opened 2019-05-02
[NIFIREG-264] add droplet type to explorer/grid-list sort and filter

scottyaslan:


bbende:
+1 looks good to me and works as expected, will merge

PR #181, opened 2019-05-03
NIFIREG-267 Updating data model to capture parameter contexts

bbende:
- Introducing v3 of flow serialization to switch to serializing VersionedFlowSnapshot instead of VersionedProcessGroup


nifi-registry-core/nifi-registry-data-model/src/main/java/org/apache/nifi/registry/flow/VersionedParameterContext.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+package org.apache.nifi.registry.flow;
+
+import io.swagger.annotations.ApiModelProperty;
+
+import java.util.Set;
+
+public class VersionedParameterContext {
+
```
kevdoran: I'm ok with this data model, but @mcgilman pointed out to me today that in NiFi Parameter Contexts will probably also have an `id` field in addition to `name` to use internally. If so, it might be worth including that in the Registry model as well.

bbende: I think since the context on the registry side is just template of the parameters (i.e. not a version controlled thing on its own), that we are ok with just a name. The NiFi side will definitely have an id so that contexts can be renamed, but that id is only relevant to the local NiFi. When you save a flow with a param context, and import to the next environment for the first time, it would create a new context with a new local id and the same name from the flow snapshot.

kevdoran:
Was able to verify that flows serialized as v2 of the flow content format can deserialize after this change. Merging now.

PR #189, opened 2019-05-15
NIFIREG-241 Added support for MySQL/MariaDB

bbende:
- Added IT tests using testcontainers project to test against Postgres, MySql, and MariaDB
- Added documentation in admin guide and README


michalklempa:
Wow. Is there a thumbs up on github?

PR #193, opened 2019-05-31
NIFIREG-276 Updating StandardFlowComparator to consider concurrent ta…

bbende:
…sks changes on public ports


bbende:
Verified this by running a full mvn clean install of this PR, then building NIFI from NIFI-6315 using -Dnifi.registry.version=0.5.0-SNAPSHOT. Make a versioned flow with public ports, after starting version control make a change to the concurrent tasks on the ports and see that there is now a local change to commit.

ijokarumawak:
LGTM +1, going to merge. Thanks @bbende !

PR #194, opened 2019-05-31
NIFIREG-212 Separating proxy into Read, Write, and Delete so some pro…

bbende:
…xies can be set to read-only

### DIFF
--- a/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/x509/X509IdentityAuthenticationProvider.java
+++ b/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/x509/X509IdentityAuthenticationProvider.java
@@ -35,12 +35,17 @@ import org.apache.nifi.registry.security.authorization.user.NiFiUserDetails;
 import org.apache.nifi.registry.security.authorization.user.StandardNiFiUser;
 import org.apache.nifi.registry.security.util.ProxiedEntitiesUtils;
 import org.apache.nifi.registry.web.security.authentication.exception.UntrustedProxyException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpMethod;
 
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Set;
 
 public class X509IdentityAuthenticationProvider extends IdentityAuthenticationProvider {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(X509IdentityAuthenticationProvider.class);
 
     private static final Authorizable PROXY_AUTHORIZABLE = new Authorizable() {
         @Override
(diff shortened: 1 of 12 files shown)


nifi-registry-core/nifi-registry-security-api/src/main/java/org/apache/nifi/registry/security/authentication/AuthenticationRequest.java
```diff
@@ -17,17 +17,20 @@
...
 
 public class AuthenticationRequest implements Serializable {
 
     private String username;
     private Object credentials;
     private Object details;
+    private String httpMethod;
 
-    public AuthenticationRequest(String username, Object credentials, Object details) {
+    public AuthenticationRequest(String username, Object credentials, Object details, String httpMethod) {
```
kevdoran: This works, but I'd prefer not to add the HttpMethod info to all instances of AuthenticationRequest as it is only every really used by the X509IdentityProvider.

The `Object details` argument of this constructor is designed to be a generic way that additional details relevant to the AuthenticationRequest can be passed from the `IdentityProvider.extractCredentials` method to the IdentityProvider implementation later for authentication. `details` can be any Object type, it is only later read by the same `IdentityProvider` instance or the AuthenticationProvider for that same IdentityProvider. In most cases, it is unused.

Currently, for the `X509IdentityProvider`, details is set to the String value of the Proxied Entities Chain Header, as that was the only relevant "detail" at the time. I would suggest just creating a new wrapper object such as X509AuthenticationRequestDetails that can hold the proxied entities chain and the HTTP method. That can be created and set in `X509IdentityProvider.extractCredentials`, and later both values can be extracted in `X509IdentityAuthenticationProvider.buildAuthenticatedToken`,

This is a small nitpick, but I'd prefer not to change AuthenticationRequest if necessary as it is part of the extension point API for third-parties, so changing this could break and third-party impls out there. I know it would also be possible to keep the current 3-arg constructor and add a 4-arg constructor, but that looses some of the benefits of this implementation such as ensuring non-null, and also adds complexity to an extension point that most people won't care about for authentication purposes. Encasing the http method in details prevents any other IdentityProviders (ours or third-parties) from needing updating, but still lets us do our authorization-as-part-of-authentication check for proxies.

bbende: Makes sense, now that I understand how the details object is being used I definitely agree that would be the better approach. Will work on updating that. Thanks!

bbende:
@kevdoran I force pushed some changes that make things work as discussed above, let me know if there is anything else, thanks!

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/security/authentication/x509/X509IdentityAuthenticationProvider.java
```diff
@@ -97,16 +106,51 @@ protected AuthenticationSuccessToken buildAuthenticatedToken(
             proxy = createUser(identity, groups, proxy, clientAddress, isAnonymous);
 
             if (chainIter.hasPrevious()) {
-                try {
-                    PROXY_AUTHORIZABLE.authorize(authorizer, RequestAction.WRITE, proxy);
-                } catch (final AccessDeniedException e) {
-                    throw new UntrustedProxyException(String.format("Untrusted proxy [%s].", identity));
+                final String httpMethodStr = x509RequestDetails.getHttpMethod().toUpperCase();
+                final HttpMethod httpMethod = HttpMethod.resolve(httpMethodStr);
+                LOGGER.debug("HTTP method is {}", new Object[]{httpMethod});
```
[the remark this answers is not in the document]
bbende: Good catch, I pushed another commit that moves it outside of the loop. 

I meant to just push up that commit but ended up rebasing so force-pushed, sorry.

scottyaslan:
@bbende thanks for this contribution. I have reviewed the front end aspects of this PR and I am a +1 to merge this in!

kevdoran:
+1 for me as well. Thanks, @bbende! Merging to master...

PR #204, opened 2019-07-04
NIFIREG-291: making NiFi Registry home directory version agnostic

ekovacs:


ekovacs:
@bbende what do you think about it?

bbende:
Looks good to me

PR #205, opened 2019-07-17
NIFIREG-285 - Add DatabaseFlowPersistenceProvider

bbende:
- Added method injection of DataSource in StandardProviderFactory
- Split out postgres migrations to use BYTEA for blob type


kevdoran:
Neat stuff! Thanks for the work on this, @bbende. It might be a while before I can get to this (I'm leaving for a trip in a couple days), but I'll definitely give it a look when I can. If you don't hear from me by the week after next, feel free to ping me again about this one.

bbende:
Thanks @kevdoran !

PR #206, opened 2019-07-25
NIFIREG-294: Added Scheduled State Changed as a difference type and u…

markap14:
…pdated the descriptors and comparators to make use of it


bbende:
+1 Looks good

PR #211, opened 2019-08-07
NIFIREG-295 Add support for proxying via Apache Knox

ijokarumawak:
- To test with Knox, apache/knox#128 is required

In order to support a reverse proxy like Knox, NiFi Registry Web UI app needs to resolve URLs relatively using the proxy server's context path. NiFi Registry will be accessed through Knox with `https://knox.example.com:8444/gateway/sandbox/nifi-registry-app/nifi-registry/`.

Currently, absolute paths are used and '/nifi-registry' is assumed to be the base path. We need to change these paths to relative ones.

### DIFF
--- a/nifi-registry-core/nifi-registry-web-ui/src/main/webapp/nf-registry-bootstrap.js
+++ b/nifi-registry-core/nifi-registry-web-ui/src/main/webapp/nf-registry-bootstrap.js
@@ -46,7 +46,7 @@ const providers = [];
 if (!locale || locale === 'en-us') {
     bootstrapModule();
 } else { //load the translation providers and bootstrap the module
-    var translationFile = './nifi-registry/assets/locale/messages.' + locale + '.xlf';
+    var translationFile = 'assets/locale/messages.' + locale + '.xlf';
 
     $.ajax({
         url: translationFile
--- a/nifi-registry-core/nifi-registry-web-ui/src/main/webapp/services/nf-registry.api.js
+++ b/nifi-registry-core/nifi-registry-web-ui/src/main/webapp/services/nf-registry.api.js
@@ -56,7 +56,7 @@ NfRegistryApi.prototype = {
      */
     getDropletSnapshotMetadata: function (dropletUri) {
         var self = this;
-        var url = '/nifi-registry-api/' + dropletUri;
+        var url = '../nifi-registry-api/' + dropletUri;
         url += '/versions';
         return this.http.get(url).pipe(
             map(function (response) {
--- a/nifi-registry-core/nifi-registry-web-ui/src/main/webapp/services/nf-registry.api.js
+++ b/nifi-registry-core/nifi-registry-web-ui/src/main/webapp/services/nf-registry.api.js
@@ -26,8 +26,8 @@ var headers = new Headers({'Content-Type': 'application/json'});
 
 var config = {
     urls: {
-        currentUser: '/nifi-registry-api/access',
-        kerberos: '/nifi-registry-api/access/token/kerberos'
+        currentUser: '../nifi-registry-api/access',
+        kerberos: '../nifi-registry-api/access/token/kerberos'
     }
 };
 
@@ -84,7 +84,7 @@ NfRegistryApi.prototype = {
      */
     getDroplet: function (bucketId, dropletType, dropletId) {
         var self = this;
-        var url = '/nifi-registry-api/buckets/' + bucketId + '/' + dropletType + '/' + dropletId;
+        var url = '../nifi-registry-api/buckets/' + bucketId + '/' + dropletType + '/' + dropletId;
         return this.http.get(url).pipe(
             map(function (response) {
                 return response;
@@ -111,7 +111,7 @@ NfRegistryApi.prototype = {
      * @returns {*}
      */
     getDroplets: function (bucketId) {
-        var url = '/nifi-registry-api/items';
+        var url = '../nifi-registry-api/items';
         if (bucketId) {
             url += '/' + bucketId;
         }
@@ -138,7 +138,7 @@ NfRegistryApi.prototype = {
      */
     deleteDroplet: function (dropletUri) {
         var self = this;
-        return this.http.delete('/nifi-registry-api/' + dropletUri, headers).pipe(
+        return this.http.delete('../nifi-registry-api/' + dropletUri, headers).pipe(
             map(function (response) {
                 return response;
             }),
@@ -162,7 +162,7 @@ NfRegistryApi.prototype = {
      */
     createBucket: function (name, allowPublicRead) {
         var self = this;
-        return this.http.post('/nifi-registry-api/buckets', {'name': name, 'allowPublicRead': allowPublicRead}, headers).pipe(
+        return this.http.post('../nifi-registry-api/buckets', {'name': name, 'allowPublicRead': allowPublicRead}, headers).pipe(
             map(function (response) {
                 return response;
             }),
@@ -186,7 +186,7 @@ NfRegistryApi.prototype = {
      */
     deleteBucket: function (bucketId) {
         var self = this;
-        return this.http.delete('/nifi-registry-api/buckets/' + bucketId, headers).pipe(
+        return this.http.delete('../nifi-registry-api/buckets/' + bucketId, headers).pipe(
             map(function (response) {
                 return response;
             }),
@@ -210,7 +210,7 @@ NfRegistryApi.prototype = {
      */
     getBucket: function (bucketId) {
         var self = this;
-        var url = '/nifi-registry-api/buckets/' + bucketId;
+        var url = '../nifi-registry-api/buckets/' + bucketId;
         return this.http.get(url).pipe(
             map(function (response) {
                 return response;
@@ -237,7 +237,7 @@ NfRegistryApi.prototype = {
      */
     getBuckets: function () {
         var self = this;
-        var url = '/nifi-registry-api/buckets';
+        var url = '../nifi-registry-api/buckets';
         return this.http.get(url).pipe(
             map(function (response) {
                 return response;
@@ -261,7 +261,7 @@ NfRegistryApi.prototype = {
      * @returns {*}
      */
     updateBucket: function (updatedBucket) {
-        return this.http.put('/nifi-registry-api/buckets/' + updatedBucket.identifier, updatedBucket, headers).pipe(
+        return this.http.put('../nifi-registry-api/buckets/' + updatedBucket.identifier, updatedBucket, headers).pipe(
             map(function (response) {
                 return response;
             }),
@@ -279,7 +279,7 @@ NfRegistryApi.prototype = {
      */
     getUser: function (userId) {
         var self = this;
-        return this.http.get('/nifi-registry-api/tenants/users/' + userId).pipe(
+        return this.http.get('../nifi-registry-api/tenants/users/' + userId).pipe(
             map(function (response) {
                 return response;
             }),
@@ -303,7 +303,7 @@ NfRegistryApi.prototype = {
      */
     addUser: function (identity) {
         var self = this;
-        return this.http.post('/nifi-registry-api/tenants/users', {
+        return this.http.post('../nifi-registry-api/tenants/users', {
             identity: identity,
             resourcePermissions: {
                 anyTopLevelResource: {
@@ -356,7 +356,7 @@ NfRegistryApi.prototype = {
      * @returns {*}
      */
     updateUser: function (identifier, identity) {
-        return this.http.put('/nifi-registry-api/tenants/users/' + identifier, {
+        return this.http.put('../nifi-registry-api/tenants/users/' + identifier, {
             'identifier': identifier,
             'identity': identity
         }, headers).pipe(
@@ -376,7 +376,7 @@ NfRegistryApi.prototype = {
      */
     getUsers: function () {
         var self = this;
-        return this.http.get('/nifi-registry-api/tenants/users').pipe(
+        return this.http.get('../nifi-registry-api/tenants/users').pipe(
             map(function (response) {
                 return response;
             }),
@@ -400,7 +400,7 @@ NfRegistryApi.prototype = {
      */
     deleteUser: function (userId) {
         var self = this;
-        return this.http.delete('/nifi-registry-api/tenants/users/' + userId, headers).pipe(
+        return this.http.delete('../nifi-registry-api/tenants/users/' + userId, headers).pipe(
             map(function (response) {
                 return response;
             }),
@@ -423,7 +423,7 @@ NfRegistryApi.prototype = {
      */
     getUserGroups: function () {
         var self = this;
-        return this.http.get('/nifi-registry-api/tenants/user-groups').pipe(
+        return this.http.get('../nifi-registry-api/tenants/user-groups').pipe(
             map(function (response) {
                 return response;
             }),
@@ -447,7 +447,7 @@ NfRegistryApi.prototype = {
      */
     getUserGroup: function (groupId) {
         var self = this;
-        return this.http.get('/nifi-registry-api/tenants/user-groups/' + groupId).pipe(
+        return this.http.get('../nifi-registry-api/tenants/user-groups/' + groupId).pipe(
             map(function (response) {
                 return response;
             }),
@@ -471,7 +471,7 @@ NfRegistryApi.prototype = {
      */
     deleteUserGroup: function (userGroupId) {
         var self = this;
-        return this.http.delete('/nifi-registry-api/tenants/user-groups/' + userGroupId, headers).pipe(
+        return this.http.delete('../nifi-registry-api/tenants/user-groups/' + userGroupId, headers).pipe(
             map(function (response) {
                 return response;
             }),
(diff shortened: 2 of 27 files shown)


ijokarumawak:
Currently, I can confirm that NiFi Registry web app is accessible through Knox. But it doesn't initialize itself as expected. It doesn't move to `explorer/grid-list`. No JS error happens.

image

By accessing NiFi Registry directly, the app is initialized successfully.
image

@rfellows do you have any advice on this behavior? (where to look at ... etc) Also, the way I changed absolute paths to relative ones look reasonable? I wonder if there's a better way. Thanks!

ijokarumawak:
I may be wrong since I have only a little experience with Angular. But it seems the difference came from [nf-registry.routes](https://github.com/apache/nifi-registry/blob/8d040bc4d66c0580e6ec21666e7771e1bce5aa12/nifi-registry-core/nifi-registry-web-ui/src/main/webapp/nf-registry.routes.js#L93). Route `pathMatch` has two mode, `full` and `prefix`, but it doesn't support `suffix` type of matching. The routes doesn't work because Knox adds `/gateway/sandbox/nifi-registry-app/` prefix to paths.

I'm going to experiment if providing a [dynamic APP_BASE_HREF](https://medium.com/@anandshende1994/setting-base-href-dynamically-in-angular-6-b7fe824848cf) can be a solution.

kevdoran:
Thanks, @ijokarumawak. Do we know if there are any backend changes required for NiFi Registry, such as URL rewriting for location URIs or anything?

ijokarumawak:
Thanks @kevdoran for checking. I haven't found anything required to change at NiFi Registry server side. Although, I noticed, NiFi Registry doesn't implement proxy context path white list as NiFi does with `nifi.web.proxy.context.path`. I'm not sure how critical it is, though. Probably we need to open another JIRA for that. How do you think?

BTW, after changing the paths to call NiFi Registry APIs to relative, the UI app started working better. I'll keep testing to see if everything works as expected.

kevdoran:
Thanks @ijokarumawak! Yes, I would agree that we should open a JIRA for adding trusted values for the various proxy headers similar to NiFi.

bbende:
@ijokarumawak one more thing to consider is that some of the entities returned by the REST API have a Link object which gets populated from LinkService. Some of the links are relative, but some are full:

https://github.com/apache/nifi-registry/blob/master/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/link/LinkService.java#L337

The baseURI comes from ApplicationResource getBaseUri() which looks at the proxy headers:

https://github.com/apache/nifi-registry/blob/master/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/ApplicationResource.java#L80-L83

So I think we just need to make sure these links are correct when proxying through Knox.

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/template.dev.html
```diff
@@ -19,7 +19,7 @@
 <html>
 <head>
     <title>NiFi Registry</title>
-    <base href="/">
+    <base href="./">
```
RynoRn: I'm not sure if this change is required. What difference it makes?

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/nf-registry-bootstrap.js
```diff
@@ -54,7 +54,7 @@ const providers = [];
 if (!locale || locale === 'en-us') {
     bootstrapModule();
 } else { //load the translation providers and bootstrap the module
-    var translationFile = './nifi-registry/assets/locale/messages.' + locale + '.xlf';
+    var translationFile = 'assets/locale/messages.' + locale + '.xlf';
```
RynoRn: I'm not sure if this was working earlier or not but with these changes this file throws an error of 404.

RynoRn:
I've looked through the changes and I have found a few things.

Please note that there are two ways to run the front-end application. Please make sure to test the application both way.

The first [1] is to run a `maven` build and then start the resulted assembly. In this case the application will start on `http://localhost:18080/nifi-registry/`.

The second way [2] is to start the backend API with the `maven` build but build and run the front-end separately from the `nifi-registry-web-ui`. In this case the front-end will pop up on `http://localhost:18081`.

Problems I found:
- Please double check if the front-end unit tests are working properly. I see **53 FAILED** test when I run `npm run test`. For example this part is probably broken: https://github.com/apache/nifi-registry/pull/211/files#diff-93f434be0ca46a3da816f4f2bc23fd04R185
- I see an error on the console if I run the app with a full build [1]
- When I refresh the application on a deep link it fails to load. E.g.: http://localhost:18081/explorer/grid-list. I'm not sure if this problem was introduced here or not. If not, feel free to ignore but would be a nice addition to fix.
- I see an error on the console when running with the full maven build [1]: `Error POST http://localhost:18080/nifi-registry-api/access/token/kerberos 409 (Conflict)`

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/template.dev.html
ijokarumawak: Knox adds a path fragment such as `/gateway/sandbox/nifi-registry-app/` in front of `/nifi-registry/`. If the `base href` is set to `/`, all relative resource references will be resolved without the Knox part.

For example, if a resource `/gateway/sandbox/nifi-registry-app/nifi-registry/images/x.png` is referenced with a relative path `images/x.png`, then `base href="/"` will resolve it as `/images/x.png` that will result a 404 error.

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/nf-registry-bootstrap.js
ijokarumawak: The only available local is `es`.
https://github.com/apache/nifi-registry/tree/master/nifi-registry-core/nifi-registry-web-ui/src/main/locale

I can get 200 result if I set browser language to Spanish. But the application failed with following error. I'll check if this was working before.
image

ijokarumawak: I've filed NIFIREG-303 and its PR is available for review #215 . It'd be great if you can review the change. Thanks!

ijokarumawak:
Current status summary:
- Deep links are should work fine now, but having warning messages details below
- I'm fixing JS tests now
- The 409 conflict error should be ignored. That has been there without this PR.
- Would you please share exact maven commands to run backend api and frontend UI separately??

The deep linking is now working as before. But I'm seeing warning messages on browser's console like below:

> Resource interpreted as Stylesheet but transferred with MIME type text/html: "https://localhost:8444/gateway/sandbox/nifi-registry-app/nifi-registry/administration/node_modules/@nifi-fds/core/common/styles/css/flow-design-system.min.css"

The master branch doesn't have this warning. Master branch returns Content-type: text/css. @RynoRn do you know what makes these difference?

ijokarumawak:
I was trying to fix the js tests. But since I'm new to js testing framework, I don't know how it works and how to debug.. @RynoRn Would you give me some resources describing how to setup debug environment or how you usually run a single test ... etc. Thanks in advance!

rfellows:
> * Would you please share exact maven commands to run backend api and frontend UI separately??

1. Build the whole nifi-registry project as normal using maven
2. Start nifi-registry as you normally do. Verify that it is running and you can access it at `http://localhost:18080/nifi-registry`
3. cd to  `nifi-registry/nifi-registry-core/nifi-registry-web-ui/src/main` folder.
4. run `npm install` to install all required js dependencies needed. BTW, this will change the `package.json` and `package-lock.json` files, just make sure to NOT check in those changes.
5. run 'npm run watch'.

The last step will compile the javascript app and fire up a web server that hosts just the UI and open a new browser tab/window with the dev-mode UI. This will be running at `http://localhost:18081/nifi-registry`. All requests for server-side resources are proxied to the running instance of nifi-registry. This will also watch the source folders for code changes and auto-rebuild as needed. When it is done it will refresh the UI in the browser.

rfellows:
> @rfellows do you have any advice on this behavior? (where to look at ... etc) Also, the way I changed absolute paths to relative ones look reasonable? I wonder if there's a better way. Thanks!

Are you still having this issue? I'm working on getting Knox set up so i can review this PR and see if I can give you any advice here. (If you have any advice on setting up Knox for this test, let me know. I've never done it.)

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/template.dev.html
rfellows: This template is only used in UI-dev-mode. I don't think one would ever configure dev mode and knox together since dev mode is just the front end and proxies requests to a running back-end.

This change breaks dev-mode when you try to navigate directly (or refresh) to `http://localhost:18081/explorer/grid-list`

I don't think this change is needed.

rfellows:
> > @rfellows do you have any advice on this behavior? (where to look at ... etc) Also, the way I changed absolute paths to relative ones look reasonable? I wonder if there's a better way. Thanks!
> 
> Are you still having this issue? I'm working on getting Knox set up so i can review this PR and see if I can give you any advice here. (If you have any advice on setting up Knox for this test, let me know. I've never done it.)

I was able to set up knox with your nifi-registry role and get it to work with nifi-registry (with your changes from this PR). When I navigate to `https://localhost:8443/gateway/sandbox/nifi-registry-app/nifi-registry` it does automatically forward me to `https://localhost:8443/gateway/sandbox/nifi-registry-app/nifi-registry/explorer/grid-list?user.name=guest` and is working. So i assume you did fix this issue that you originally asked me about? Or did I misunderstand the question?

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/services/nf-registry.api.js
```diff
@@ -82,7 +82,7 @@ NfRegistryApi.prototype = {
      */
     getDroplet: function (bucketId, dropletType, dropletId) {
         var self = this;
-        var url = '/nifi-registry-api/buckets/' + bucketId + '/' + dropletType + '/' + dropletId;
+        var url = '../nifi-registry-api/buckets/' + bucketId + '/' + dropletType + '/' + dropletId;
```
rfellows: Looks like you missed one on line 58 (`getDropletSnapshotMetadata` still points to `/nifi-registry-api/`). 
Github doesn't let me comment on lines that were not edited so the comment is here.

ijokarumawak:
Thank you very much @rfellows for your comments! I have:
- Fixed all broken JS tests
- Confirmed dev mode UI works

It'd be great if the commands to start UI dev mode is written on a doc such as README or Wiki page.

Registry works fine behind Knox. Deeplinks work fine, too. But still seeing following warning messages on browser console.

```
Resource interpreted as Stylesheet but transferred with MIME type text/html: "https://localhost:8444/gateway/sandbox/nifi-registry-app/nifi-registry/administration/node_modules/@nifi-fds/core/common/styles/css/flow-design-system.min.css".
```

rfellows:
> Registry works fine behind Knox. Deeplinks work fine, too. But still seeing following warning messages on browser console.
> 
> ```
> Resource interpreted as Stylesheet but transferred with MIME type text/html: "https://localhost:8444/gateway/sandbox/nifi-registry-app/nifi-registry/administration/node_modules/@nifi-fds/core/common/styles/css/flow-design-system.min.css".
> ```

As I mentioned in the comment above, this should go away once #209 is merged. This PR will have to be rebased on top of that (with some light merging I expect).

mcgilman:
#209 has been merged. This should alleviate the js test failures that were being discussed here.

ijokarumawak:
Rebased with the latest master. I think I understand why I'm still seeing the warning message.

> Resource interpreted as Stylesheet but transferred with MIME type text/html: "https://nifi-registry:18443/nifi-registry/administration/vendor.min.14dc6f3729ebc0fec9e2.css".

When NiFi Registry Web UI is accessed from a browser, following HTML is returned (which is derived from template.html):

```html
<html>
<head>
    <title>NiFi Registry</title>
    <script type="text/javascript">
        var pathname = document.location.pathname;
        var appRootIndex = document.location.pathname.indexOf('/nifi-registry/');
        var base = document.createElement('base');
        base.setAttribute('href', document.location.pathname.substring(0, appRootIndex) + '/nifi-registry/');
        document.head.appendChild(base);
    </script>
    <meta charset='UTF-8'>
    <meta name='viewport' content='width=device-width, initial-scale=1'>
    <meta http-equiv='Content-Type' content='text/html; charset=UTF-8'/>
    <!-- Styles will be injected here by webpack -->
<link rel="shortcut icon" href="registry-favicon.png"><link href="vendor.min.14dc6f3729ebc0fec9e2.css" rel="stylesheet"><link href="nf-registry.style.min.2bc4be4717d8b764b9ba.css" rel="stylesheet"></head>
<body>
    <nf-registry-app></nf-registry-app>
    <!-- Scripts will be injected here by webpack -->
<script type="text/javascript" src="vendor.min.e33e47d8e9966542cb4a.js"></script><script type="text/javascript" src="nf-registry.bundle.min.236c34497ded3ce25502.js"></script></body>
</html>
```

Then, browsers (I tested Chrome and FireFox) do followings:
1. Loads external resources, CSS and JS in this case, using the `href` that is a relative link from current URL. For example, if user hard reloads (Shift+Ctrl+R) `https://nifi-registry:18443/nifi-registry/explorer/grid-list`, browser sends a GET request to `https://nifi-registry:18443/nifi-registry/explorer/grid-list/vendor.min.14dc6f3729ebc0fec9e2.css`. But the resource is not found, so HTML is returned, then browser logs the warning message.
2. After that happens, the script block is evaluated, and adds `<base href="/nifi-registry/">`.
3. Browser reloads external resources once again. This time using `https://nifi-registry:18443/nifi-registry/vendor.min.14dc6f3729ebc0fec9e2.css` successfully. I confirmed 2 GET requests sent for each CSS.

@rfellows @mcgilman Do you have any solution to handle this? A possible one would be changing how webpack inject the resource links. For example, injecting a script block that will add 'link' and 'script' tags dynamically after the 'base' tag is added, so that the relative paths are evaluated correctly. I don't know where to change to do that, though...  How do you think?

ijokarumawak:
- @kevdoran I filed NIFIREG-306 to implement context path verification
- @bbende Currently ExtensionRepoResource is the only resource using LinkService.populateFullLinks. I confirmed nifi-registry-api/extension-repository uses proxy related HTTP headers to generate full URLs as expected.

Now, the remaining thing for this PR to be closed is the warning messages when browser load CSS/JS before '<head>' tag is generated. Please look at [this comment](https://github.com/apache/nifi-registry/pull/211#issuecomment-521483420) for detail.

rfellows:
> @rfellows @mcgilman Do you have any solution to handle this? A possible one would be changing how webpack inject the resource links. For example, injecting a script block that will add 'link' and 'script' tags dynamically after the 'base' tag is added, so that the relative paths are evaluated correctly. I don't know where to change to do that, though... How do you think?

@ijokarumawak I spent some time working this and came up with an alternative approach. It basically uses URL fragments rather than the entire URL... making everything work from the same "base" location.

I added 2 commits on top of yours in my own fork (but i rebased your work on to the latest from master first):
https://github.com/ijokarumawak/nifi-registry/compare/nifireg-295...rfellows:pr211?expand=1

So, if you rebase your work on apache master, you should be able to cherry-pick my last 2 commits from my branch (https://github.com/rfellows/nifi-registry/tree/pr211) and have it work for you.

I tested normal mode, dev mode, and through knox. I was not able to successfully test a secured version of registry through knox however. I am not sure it that is supported or not and if it is, how to configure it properly.

Let me know if you have any questions.

scottyaslan:
Thanks for digging into this @rfellows. I am +1 for this approach you have outlined.

ijokarumawak:
Thank you very much @rfellows ! Using hash makes everything clean! I'm also +1 with that approach. Rebased with the latest master and cherry-picked @rfellows commits.

Can we merge this PR now?

rfellows:
> Thank you very much @rfellows ! Using hash makes everything clean! I'm also +1 with that approach. Rebased with the latest master and cherry-picked @rfellows commits.
> 
> Can we merge this PR now?

@ijokarumawak should knox with a secured registry work? I would except yes. But I couldn’t get it to work. What is the missing piece?

ijokarumawak:
@rfellows I tested with a secured NiFi Registry behind Knox and it works. You may find this document helpful. You need to setup keystores on both Knox and NiFi Registry so that they can trust each other. https://docs.hortonworks.com/HDPDocuments/HDF3/HDF-3.4.1.1/nifi-knox/content/configuring_knox-for-nifi.html

rfellows:
@ijokarumawak Thanks! I knew there had to be something I was missing.

RynoRn:
Thanks for helping in @rfellows! The suggested approach looks great. 👍

scottyaslan:
I have also reviewed this PR and the changes look good. I built it and ran a secured (with Certs) local build and verified the deep links and api calls still work. +1

bbende:
Looks like there is consensus that this is good to go, thanks @ijokarumawak !! Will merge shortly

PR #212, opened 2019-08-09
NIFIREG-300 Added nifi-registry-revision modules containg NiFi's Revi…

bbende:
…sionManager concept with a JDBC implementation and supporting utility modules

This pull request doesn't change any functionality in the application, it just establishes the nifi-registry-revision modules. It is probably easiest to review them by looking through the modules in an IDE to see how the pieces fit together, starting with nifi-registry-revision-api which directly correlates to NiFi's revision API (potentially they could share this API module once released).

### DIFF
--- a/nifi-registry-core/nifi-registry-revision/nifi-registry-revision-api/src/main/java/org/apache/nifi/registry/revision/api/Revision.java
+++ b/nifi-registry-core/nifi-registry-revision/nifi-registry-revision-api/src/main/java/org/apache/nifi/registry/revision/api/Revision.java
@@ -16,6 +16,12 @@
  */
 package org.apache.nifi.registry.revision.api;
 
+/**
+ * A revision for an entity which is made up of the entity id, a version, and an optional client id.
+ *
+ * NOTE: This API is considered a framework level API for the NiFi ecosystem and may evolve as
+ * the NiFi PMC and committers deem necessary. It is not considered a public extension point.
+ */
 public class Revision {
 
     private final Long version;
--- a/nifi-registry-core/nifi-registry-revision/nifi-registry-revision-api/src/main/java/org/apache/nifi/registry/revision/api/Revision.java
+++ b/nifi-registry-core/nifi-registry-revision/nifi-registry-revision-api/src/main/java/org/apache/nifi/registry/revision/api/Revision.java
@@ -1,0 +1,111 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.registry.revision.api;
+
+public class Revision {
+
+    private final Long version;
+    private final String clientId;
+    private final String entityId;
+
+    /**
+     * @param version the version number for the revision
+     * @param clientId the id of the client creating the revision, or null if one is not provided
+     * @param entityId the id of the component the revision belongs to
+     */
+    public Revision(final Long version, final String clientId, final String entityId) {
+        if (version == null) {
+            throw new IllegalArgumentException("The revision must be specified.");
+        }
+        if (entityId == null) {
+            throw new IllegalArgumentException("The entityId must be specified.");
+        }
+
+        this.version = version;
+        this.clientId = clientId;
+        this.entityId = entityId;
+    }
+
+    public String getClientId() {
+        return clientId;
+    }
+
+    public Long getVersion() {
+        return version;
+    }
+
+    public String getEntityId() {
+        return entityId;
+    }
+
+    /**
+     * Returns a new Revision that has the same Client ID and Component ID as this one, but with a larger version.
+     *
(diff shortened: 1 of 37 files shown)


pom.xml
```diff
@@ -94,13 +94,14 @@
...
         <jax.rs.api.version>2.1</jax.rs.api.version>
         <jersey.version>2.27</jersey.version>
-        <jackson.version>2.9.8</jackson.version>
-        <spring.boot.version>2.1.3.RELEASE</spring.boot.version>
-        <spring.security.version>5.1.3.RELEASE</spring.security.version>
-        <flyway.version>5.2.1</flyway.version>
+        <jackson.version>2.9.9</jackson.version>
+        <spring.boot.version>2.1.6.RELEASE</spring.boot.version>
+        <spring.security.version>5.1.5.RELEASE</spring.security.version>
+        <flyway.version>5.2.4</flyway.version>
```
kevdoran: Do any of these dep version changes require NOTICE file changes? not sure if we mention version numbers there or not.

nifi-registry-core/nifi-registry-revision/nifi-registry-revision-spring-jdbc/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+        <dependency>
+            <groupId>org.apache.nifi.registry</groupId>
+            <artifactId>nifi-registry-revision-common</artifactId>
+            <version>0.5.0-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-jdbc</artifactId>
+            <version>${spring.boot.version}</version>
+        </dependency>
```
kevdoran: Is `spring-boot-starter-jdbc`necessary? Unless I'm missing something it seems like only `spring-jdbc` is needed.

nifi-registry-core/nifi-registry-revision/nifi-registry-revision-entity-model/src/main/java/org/apache/nifi/registry/revision/entity/RevisionInfo.java
```diff
@@ -0,0 +1,79 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.nifi.registry.revision.entity;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel(description = "The revision information for an entity managed through the REST API.")
+public class RevisionInfo {
```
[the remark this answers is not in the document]
kevdoran: Disregard this comment. I see the classes are different now.

nifi-registry-core/nifi-registry-revision/nifi-registry-revision-api/src/main/java/org/apache/nifi/registry/revision/api/Revision.java
```diff
@@ -0,0 +1,111 @@
...
+            return false;
+        }
+        if (thatRevision.getEntityId() != null && !thatRevision.getEntityId().equals(getEntityId())) {
+            return false;
+        }
+
+        if (this.version != null && this.version.equals(thatRevision.version)) {
+            return true;
+        } else {
+            return clientId != null && !clientId.trim().isEmpty() && clientId.equals(thatRevision.getClientId());
```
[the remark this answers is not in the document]
bbende: I think you are right that the trim isn't helpful here, this was existing NiFi code that was ported so I will probably leave it as is until we know of an issue that requires changing it

nifi-registry-core/nifi-registry-revision/nifi-registry-revision-common/src/main/java/org/apache/nifi/registry/revision/standard/RevisionComparator.java
```diff
@@ -0,0 +1,40 @@
...
+public class RevisionComparator implements Comparator<Revision> {
+
+    @Override
+    public int compare(final Revision o1, final Revision o2) {
+        final int entityComparison = o1.getEntityId().compareTo(o2.getEntityId());
+        if (entityComparison != 0) {
+            return entityComparison;
+        }
+
+        final int clientComparison = o1.getClientId().compareTo(o2.getClientId());
```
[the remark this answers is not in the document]
bbende: Good catch, I think the reason this ends up never happening is because in the "clientId" parameter class in the REST layer, we have:

```
public ClientIdParameter(String clientId) {
        if (clientId == null || clientId.trim().isEmpty()) {
            this.clientId = UUID.randomUUID().toString();
        } else {
            this.clientId = clientId;
        }
    }
```
So it always ends up non-null, but I'm going to update this code with something like this:
```

final Comparator<String> nullSafeStringComparator = 
        Comparator.nullsFirst(String::compareTo);

final int clientComparison = Objects.compare(
        o1.getClientId(), o2.getClientId(), nullSafeStringComparator);
```

pom.xml
bbende: I will double check, but they only would require changes if they brought in need dependencies that also required mentions in the NOTICE, or I guess if the copyright years changed.

nifi-registry-core/nifi-registry-revision/nifi-registry-revision-spring-jdbc/pom.xml
bbende: Good point, I'll try using just spring-jdbc and see what happens.

The only part that makes me hesitant is then we basically have to maintain a version of spring-jdbc, independently of the version of spring-boot.

bbende:
@kevdoran pushed a new commit that addresses the comments (comparator null check, NOTICE updates, use of spring-jdbc). 

I also added a note in the javadoc of all the API classes stating this is really a framework level API and not a public extension point, so we are free to evolve as necessary.

Let me know if there is anything else.

nifi-registry-core/nifi-registry-revision/nifi-registry-revision-spring-jdbc/pom.xml
bbende: Seems to work fine just using spring-jdbc so I'll leave it with that.

PR #220, opened 2019-08-19
NIFIREG-310: Support case insensitive membership decisions in LdapUserGroupProvider

mcgilman:
NIFIREG-310:
- Introducing case insensitive group membership.
- Ensuring user identity and group name mapping is applied when the user identity or group name is inferred through group membership decisions.


bbende:
Will review...

PR #231, opened 2019-09-25
NIFIREG-320 - Added jetty header filters to set security headers. Set…

thenatog:
…ting security headers for the registry-api using spring security configuration.

NIFIREG-320 - Fixed rest-api docs loading. Headers confirmed to be applied for docs.

### DIFF
--- a/nifi-registry-core/nifi-registry-jetty/src/main/java/org/apache/nifi/registry/jetty/JettyServer.java
+++ b/nifi-registry-core/nifi-registry-jetty/src/main/java/org/apache/nifi/registry/jetty/JettyServer.java
@@ -17,6 +17,10 @@
 package org.apache.nifi.registry.jetty;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.nifi.registry.jetty.headers.ContentSecurityPolicyFilter;
+import org.apache.nifi.registry.jetty.headers.StrictTransportSecurityFilter;
+import org.apache.nifi.registry.jetty.headers.XFrameOptionsFilter;
+import org.apache.nifi.registry.jetty.headers.XSSProtectionFilter;
 import org.apache.nifi.registry.properties.NiFiRegistryProperties;
 import org.apache.nifi.registry.security.crypto.CryptoKeyProvider;
 import org.eclipse.jetty.annotations.AnnotationConfiguration;
@@ -28,11 +32,10 @@ import org.eclipse.jetty.server.SecureRequestCustomizer;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.SslConnectionFactory;
-import org.eclipse.jetty.server.handler.ContextHandler;
 import org.eclipse.jetty.server.handler.HandlerCollection;
-import org.eclipse.jetty.server.handler.ResourceHandler;
-import org.eclipse.jetty.util.resource.Resource;
-import org.eclipse.jetty.util.resource.ResourceCollection;
+import org.eclipse.jetty.servlet.DefaultServlet;
+import org.eclipse.jetty.servlet.FilterHolder;
+import org.eclipse.jetty.servlet.ServletHolder;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
 import org.eclipse.jetty.webapp.Configuration;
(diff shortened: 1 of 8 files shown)


nifi-registry-core/nifi-registry-web-api/pom.xml
```diff
@@ -445,5 +445,11 @@
             <version>${jetty.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.nifi.registry</groupId>
+            <artifactId>nifi-registry-jetty</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
```
bbende: We shouldn't have to add this since the web-api uses Spring Security's configuration and doesn't use our custom filters right?

nifi-registry-core/nifi-registry-jetty/src/main/java/org/apache/nifi/registry/jetty/JettyServer.java
```diff
@@ -301,6 +338,15 @@ private WebAppContext loadWar(final File warFile, final String contextPath, fina
...
         webappContext.setMaxFormContentSize(600000);
 
+        // add HTTP security headers to all responses
+        final String ALL_PATHS = "/*";
+        ArrayList<Class<? extends Filter>> filters = new ArrayList<>(Arrays.asList(XFrameOptionsFilter.class, ContentSecurityPolicyFilter.class, XSSProtectionFilter.class));
+        if(properties.isHTTPSConfigured()) {
+            filters.add(StrictTransportSecurityFilter.class);
+        }
+
+        filters.forEach( (filter) -> addFilters(filter, ALL_PATHS, webappContext));
```
bbende: Should we be somehow skipping over the webApiContext since we are handling that a different way?

I suppose it doesn't really matter, but I think we end up with the web-api having both sets of filters, but never reaching these since they end up after Jersey.

nifi-registry-core/nifi-registry-jetty/src/main/java/org/apache/nifi/registry/jetty/JettyServer.java
```diff
@@ -106,6 +112,37 @@ public JettyServer(final NiFiRegistryProperties properties, final CryptoKeyProvi
...
+     * exist an attempt will be made to create the directory. If that is
+     * successful NIFI will no longer fail and will start successfully barring
+     * any other errors. The side effect of the docsDirectory not being present
+     * is that the documentation links under the 'General' portion of the help
+     * page will not be accessible, but at least the process will be running.
+     *
+     * @param docsDirectory Name of documentation directory in installation directory.
+     * @return A File object to the documentation directory; else startUpFailure called.
+     */
+    private File getDocsDir(final String docsDirectory) {
```
bbende: I'm wondering if this change was really needed... looking at lines 383-394 of the original code, it looks like the logic there was basically the same that is in this method.

Was there anything else that was being fixed by changing the way the docs webapp was being loaded?

nifi-registry-core/nifi-registry-web-api/pom.xml
thenatog: Yep I'll remove that

nifi-registry-core/nifi-registry-jetty/src/main/java/org/apache/nifi/registry/jetty/JettyServer.java
thenatog: The main thing here to change the way the docs are loaded was by using a ServletHolder instead of a ResourceHandler. The getDocsDir method was extracted in NiFi to this separate method, so I basically just reused that.

nifi-registry-core/nifi-registry-jetty/src/main/java/org/apache/nifi/registry/jetty/JettyServer.java
thenatog: I didn't check what the context looks like after using the spring security settings, but it doesn't appear to affect much using both. I can skip over it if you think that's better.

nifi-registry-core/nifi-registry-jetty/src/main/java/org/apache/nifi/registry/jetty/JettyServer.java
bbende: I guess my question was... did we need to change the way the docs webapp was loaded in order for something with the security headers to work correctly? 

I'm only asking since I don't think there was any existing issue with the docs webapp.

thenatog: Yes, without using a ServletHolder for the docs webapp, the Jetty filters and thereby the security headers are not applied correctly. This is something I experienced when attempting the change in NiFi. Basically when using a ResourceHandler, when the request is made to the docs the filters are not triggered. I did research into this, possibly emailed the Jetty dev distro, and eventually found that I had to use a ServletHolder.

bbende:
Looks good, will merge

PR #232, opened 2019-09-26
NIFIREG-292 Add DB impls of UserGroupProvider and AccessPolicyProvider

bbende:
This will allow us to get all data (metadata, flow content, and users/groups/policies) into the DB, and therefore can run more than one instance of the application pointing to the same external DB.

This PR is based off the other PR for NIFIREG-321, so that needs to be reviewed and merged first.

There is also a bunch of refactoring to the file-based providers so we should also test existing functionality with the file-based providers.


kevdoran:
@bbende I hope to get to this PR this week. Can you please rebase when you get a chance?

bbende:
@kevdoran thanks! yup already working on it, needs a little refactoring based on the PR I merged this morning to support the 'NiFI Group Name' concept.

kevdoran:
I had paused review work on this until #230 was finalized, but it looks like that one is fairly close and any pending changes there should be minor, so I'll pick this one back up today.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/database/DatabaseAccessPolicyProvider.java
```diff
@@ -0,0 +1,122 @@
...
+            LOGGER.info("Populating authorizations for NiFi identities: [{}]", StringUtils.join(nifiIdentities, ";"));
+            populateNiFiIdentities(nifiIdentities);
+        }
+
+        if (!StringUtils.isBlank(nifiGroupName)) {
+            LOGGER.info("Populating authorizations for NiFi Group: '" + nifiGroupName + "'");
+            populateNiFiGroup(nifiGroupName);
+        }
+    }
+
```
kevdoran: So, a few general thoughts that apply to both `AbstractDatabaseAPP->DatabaseAPP` and also `AbstractDatabaseUGP->DatabaseUGP`.

Recently, I've personally been trying to avoid two things in my own code:

1. Static utility classes that do anything more than trivial checks/transforms
2. The pattern of abstract base classes that delegate required functionality to abstract protected methods. It works, but I just find it harder to understand, test, maintain, and reuse. (That might just be me!)

An alternative that usually works for me is making the base class concrete and defining an interface type (or multiple interfaces) that can be injected into that class.

So using DatabaseAccessPolicyProvider as an example, the concrete subclass could be collapsed into the base class, which would no longer be abstract. The class could then be injected with the DataSource and also a new interface, something like this:

```java
interface IdentityMappingProvider {
    List<IdentityMapping> getIdentityMappings()
}
```

or perhaps even better:

```java
interface IdentityMapper {
    String mapIdentity(String rawIdentity)
}
```

In NiFiRegistry, we could create implementations of `IdentityMappingProvider` or `IdentityMapper` that are backed by a properties object or NiFiRegistryProperties specifically, but if we wanted to reuse these DatabaseAPP and DatabaseUGP classes in other apps, it is easier to reuse them as those apps would merely need to provide implementations that make sense in those contexts. These new interfaces would do away with the need for IdentityMappingUtil (as the logic would move to those implementing classes), as well as any location in our code that is directly coupled to NiFiRegsitryProperties solely for the purpose of passing to IdentityMappingUtil.

Now in this case, constructor injection / autowiring does not work because of the way the classes are created by the framework using JAX-B and the extension/provider interface we are supporting. (Side note: I think we should revisit the way we support this as part of 1.0 or 2.0, but that is definitely outside the scope of this PR). However, I do think extending the `AuthorizerConfigurationContext` is on the table. For example:

```java
public interface AuthorizerConfigurationContext {
    // existing methods
    String getIdentifier();
    Map<String, String> getProperties();
    PropertyValue getProperty(String property);

    // new methods
    DataSource getDataSource();
    IdentityMapper getIdentityMapper();
}
```

In this way, the implementations of the interfaces could be injected into `AuthorizerFactory` and passed along to extensions when `AuthorizerFactory` creates the configuration context and calls `onConfigured(context)`. Provider implementations could use those objects if needed or ignore them if not applicable to the provider impl.

Let me know what you think. I know this is done/working and the changes I propose are non-trivial. I am happy with any of the following:
- Leaving as is if you prefer the approach used here
- Doing this refactoring later, perhaps a part of a larger set of changes (for example, there are lot of places that NiFiRegistry injections and the existing IdentityMappingUtil is used, and my suggestions here apply to those existing locations as well).
- Providing some of the work that would make this easier, but contributing PRs to your branch or to master directly then we could then use to rebase this branch upon and take advantage of new interfaces/interface methods.

bbende: These are some good points. I think you are right that this one of those cases where you can say "composition over inheritance", since the abstract class wasn't really for polymorphism.

I like the idea of the `IdentityMapper` interface and I think we can inject that into the authorizer from the factory the same we are injecting the `DataSource` and `NiFiRegistryProperties` currently (I agree on possibly changing how that works in the future).

I would rather do the custom injection in the factory rather then add the methods to the `AuthorizerConfigurationContext`, only because the context class is part of the API and used across all impls, and an application may not have a `DataSource` to provide. The `IdentityMapper` could make sense there since that interface could be part of the API, but it also blurs the lines a little on what the config context is, currently it represents the properties from the config file.

Long story short, I'll work on some updates around the suggestions here.

nifi-registry-core/nifi-registry-framework/src/main/resources/db/migration/default/V8__AddUserGroupPolicy.sql
```diff
@@ -0,0 +1,62 @@
...
+    IDENTITY VARCHAR(4096) NOT NULL,
+    CONSTRAINT PK__UGP_GROUP_IDENTIFIER PRIMARY KEY (IDENTIFIER),
+    CONSTRAINT UNIQUE__UGP_GROUP_IDENTITY UNIQUE (IDENTITY)
+);
+
+CREATE TABLE UGP_USER_GROUP (
+    USER_IDENTIFIER VARCHAR(50) NOT NULL,
+    GROUP_IDENTIFIER VARCHAR(50) NOT NULL,
+    CONSTRAINT PK__UGP_USER_GROUP PRIMARY KEY (USER_IDENTIFIER, GROUP_IDENTIFIER),
+    CONSTRAINT FK__UGP_USER_GROUP_USER_IDENTIFIER FOREIGN KEY (USER_IDENTIFIER) REFERENCES UGP_USER(IDENTIFIER) ON DELETE CASCADE,
```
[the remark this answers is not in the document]
kevdoran: Yep it is possible in the file UGP.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/database/DatabaseAccessPolicyProvider.java
kevdoran: Cool that all sounds good. Good point about the DataSource not always being available depending on the app/context.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/AbstractPolicyBasedAuthorizer.java
```diff
@@ -269,10 +272,13 @@ public final synchronized User updateUser(final User user) throws AuthorizationA
     /**
      * Deletes the user with the given id.
      *
+     * @deprecated use {@link #deleteUser(User)} instead
+     *
      * @param userIdentifier the identifier of the user to delete
      * @return the user that was deleted, or null if no matching user was found
      * @throws AuthorizationAccessException if there was an unexpected error performing the operation
      */
+    @Deprecated
```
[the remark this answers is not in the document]
kevdoran: That makes sense, thanks for the explanation. I guess I would lean towards either of these options, but don't have a preference between the two:

- Remove the deprecated notice, the methods will just be unused
- Remove them all together in this PR since it is a major version anyway

bbende:
@kevdoran pushed up some new commits that address the feedback we discussed above..

- Removed unused deprecated methods
- Removed FK constraint on user-group table
- Introduced IdentityMapper interface and integrated into various providers
- Collapsed abstract DB provider into impls

For now the DefaultIdentityMapper is just a wrapper around IdentityMapperUtil just so we can get all the direct references to IdentityMappingUtil and NiFiRegistryProperties out of the providers, but we can improve on this default impl later and probably get rid of the util class all together.

kevdoran:
Sounds good, thanks @bbende! will try to re-review today or tomorrow.

bbende:
@kevdoran this PR has been rebased against master and should now contain only the changes for NIFIREG-292. Let me know if there are any issues.

kevdoran:
@bbende Thanks for keeping this PR up to date. I plan to review it this week.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/util/UserGroupProviderUtils.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public static final String PROP_INITIAL_USER_IDENTITY_PREFIX = "Initial User Identity ";
+    public static final Pattern INITIAL_USER_IDENTITY_PATTERN = Pattern.compile(PROP_INITIAL_USER_IDENTITY_PREFIX + "\\S+");
+
+    public static Set<String> getInitialUserIdentities(final AuthorizerConfigurationContext configurationContext, final IdentityMapper identityMapper) {
+        final Set<String> initialUserIdentities = new HashSet<>();
+        for (Map.Entry<String,String> entry : configurationContext.getProperties().entrySet()) {
+            Matcher matcher = UserGroupProviderUtils.INITIAL_USER_IDENTITY_PATTERN.matcher(entry.getKey());
+            if (matcher.matches() && !StringUtils.isBlank(entry.getValue())) {
+                initialUserIdentities.add(identityMapper.mapUser(entry.getValue()));
```
kevdoran, line 42: Same question here as my comment in APPUtils regarding the use of IdentityMapper to map user-provided values.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/file/FileAccessPolicyProvider.java
```diff
@@ -70,17 +69,13 @@
...
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
-public class FileAccessPolicyProvider implements ConfigurableAccessPolicyProvider {
+public class FileAccessPolicyProvider extends AbstractConfigurableAccessPolicyProvider {
```
kevdoran, line 78: wow this class got a lot simpler 👍

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/database/DatabaseUserGroupProvider.java
```diff
@@ -0,0 +1,387 @@
...
+            final String userGroupSql =
+                    "SELECT " +
+                            "G.IDENTIFIER AS IDENTIFIER, " +
+                            "G.IDENTITY AS IDENTITY " +
+                    "FROM " +
+                            "UGP_GROUP AS G, " +
+                            "UGP_USER_GROUP AS UG " +
+                    "WHERE " +
+                            "G.IDENTIFIER = UG.GROUP_IDENTIFIER AND " +
+                            "UG.USER_IDENTIFIER = ?";
```
kevdoran, line 203: Thanks for maintaining al these JDBC templates 🙂

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/util/AccessPolicyProviderUtils.java
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Returns the value of the 'Initial Admin Identity' property with any identity mappings applied.
+     *
+     * @param configurationContext the configuration context
+     * @param identityMapper the identity mapper
+     * @return the value for the initial admin identity
+     */
+    public static String getInitialAdminIdentity(final AuthorizerConfigurationContext configurationContext, final IdentityMapper identityMapper) {
+        final PropertyValue initialAdminIdentityProp = configurationContext.getProperty(PROP_INITIAL_ADMIN_IDENTITY);
+        return initialAdminIdentityProp.isSet() ? identityMapper.mapUser(initialAdminIdentityProp.getValue()) : null;
```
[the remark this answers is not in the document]
bbende: The existing file-based implementations in Nifi and Nifi registry both apply the identity mappings inside the user-group-provider and access-policy-provider, so this was mostly preserving existing behavior and making it reusable across the file and DB impls.

It basically means you could enter either form... entering pre-mapped would map it for you, or if you know the mapping result you could enter that and it won't match the patterns anymore. 

It is somewhat questionable that each implementation has to do this separately though. Maybe there is a way we could move it to a layer above as part of a follow on effort, but I haven't thought enough about it, or we could not apply it in any of them and you always have to enter post-mapped values.

bbende:
Thanks for reviewing! I agree that an IT test would be really nice, especially given the test container DB profiles you mentioned. I think we already have a similar example with an IT test for the DB flow persistence provider. Happy to work on that if you make a follow up JIRA, or if there ends up being any other additional work to close this out then could add it here.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/util/AccessPolicyProviderUtils.java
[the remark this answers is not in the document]
kevdoran: Yeah that makes sense. Thanks for clarifying that this preserves existing behavior. I checked but had missed that. This is good for me.

PR #233, opened 2019-09-30
NIFIREG-323 Clear secret key when auto restarting in order to obtain …

bbende:
…new secret key from the NiFi Registry process


thenatog:
Reviewing..

thenatog:
Works as expected. I tested that when killing the nifi-registry process (not the bootstrap process), the daemon does restart nifi-registry but bootstrap commands like ./nifi-registry.sh [status|stop] do not work. Once applying this fix, the restarted process will now communicate the new key back to the bootstrap daemon and the above commands will continue to work on the newly spawned process.

PR #235, opened 2019-10-03
NIFIREG-325: support for node identity group

ekovacs:


bbende:
Will review..

bbende:
We should update the default authorizers.xml similar to this change in NiFi:

https://github.com/apache/nifi/pull/2970/files#diff-8f577f792e128744ad8f8f5eea585ac4

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/file/FileAccessPolicyProvider.java
```diff
@@ -147,6 +149,7 @@ private static JAXBContext initializeJaxbContext(final String contextPath) {
 
     static final String PROP_NIFI_IDENTITY_PREFIX = "NiFi Identity ";
     static final String PROP_USER_GROUP_PROVIDER = "User Group Provider";
+    static final String PROP_NIFI_IDENTITY_GROUP_NAME = "Identity Group Name";
```
bbende: I think we should probably call this `NiFi Identity Group Name` since that would correspond with the other properties for `NiFi Identity`

ekovacs: Initially i was going to call it `NiFi Identity Group` or `NiFi Identity Group Name` but then i realised that:
if no node identites are present, only NiFi Identity Group Name, then the `PROP_NIFI_IDENTITY_PREFIX` would regex match it, and would create user authorizations for it. Thus i opted to name it differently to avoid false instance `NiFi Identity` matches

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/file/FileAccessPolicyProvider.java
```diff
@@ -215,6 +219,23 @@ public void onConfigured(AuthorizerConfigurationContext configurationContext) th
                 }
             }
 
+            PropertyValue identityGroupNameProp = configurationContext.getProperty(PROP_NIFI_IDENTITY_GROUP_NAME);
+            String identityGroupName = (identityGroupNameProp != null && identityGroupNameProp.isSet()) ? identityGroupNameProp.getValue() : null;
+            if (identityGroupName != null) {
```
bbende: I haven't dug into the difference yet, but in NiFi if you have `<property name="Node Group"></property>`, it doesn't look for the group and continues on with normal start up, but in registry if you have `<property name="Identity Group Name"></property>` then it throws an exception:

`SecurityProviderCreationException: Authorizations node group '' could not be found`

We should probably try to have the same behavior here.

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/file/FileAccessPolicyProvider.java
bbende: Ok that makes sense, how about just `NiFi Group Name` then?

ekovacs:
Thank you @bbende for your review. Just updated the PR with your suggestions.

PR #240, opened 2019-10-16
NIFIREG-329 Add build profile that tests all DBs

kevdoran:
Note, this PR builds on #239, so that should be merged first and this rebased after that.


bbende:
This is really helpful! Looks good, will merge

PR #243, opened 2019-10-18
NIFIREG-338: fixing version in dockerfile

ekovacs:


ekovacs:
@bbende could you please take a look at this tiny change?

PR #245, opened 2019-10-18
NIFIREG-252: adding mavendocker build profile

ekovacs:


ekovacs:
@kevdoran could you please take a look? / Give it a run?

ekovacs:
`mvn clean  package verify -P include-ranger -P include-aws -P docker`

creates the docker image locally

kevdoran:
reviewing...

nifi-registry-docker-maven/dockermaven/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+                                <configuration>
+                                    <arguments>
+                                        <argument>${project.version}-dockermaven</argument>
+                                        <argument>${project.version}</argument>
+                                    </arguments>
+                                    <executable>${project.basedir}/integration-test.sh</executable>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
```
kevdoran: We should disable executing this plugin when `-DskipTests` is used with the mvn build.

(It occurs to me that this would be a nice enhancement for nifi as well, which uses a similar mechanism for testing the dockermaven result.)

ekovacs:
> Thanks for the contribution @ekovacs!
> 
> Overall this looks good to me.
> 
> I had one inline suggestion about disabling a plugin execution under certain conditions, but that is more of a nice-to-have than a deal-breaker.
> 
> At some point I would like to try to combine this with the `nifi-registry-docker` module, especially to de-duplicate items such as the shell scripts that get added to the image, but we can save that for a future improvement to the project.
> 
> Ran the build and the output is working for me. This will help with testing, esp. integration tests with nifi.
> 
> Thanks!

at first, i put `dockermaven` project as a child of nifi-registry-docker (which is nifi-registry-core's child), but then i was unable to make sure that `nifi-registry-assembly` would execute prior to `dockermaven` (because `nifi-registry-core` must be built prior to `nifi-registry-assembly`), 

so i put it under the root directly, and specified `nifi-registry-docker-maven` later then `nifi-registry-assembly` in the root `pom.xml`'s module list, so this way i have the built artifacts ready when starting the docker build.

i could have moved the existing `nifi-registry-docker` to the root to have docker things in one place, but i didn't want to break any external infra which may expect `nifi-registry-docker` at the currrent location

ekovacs:
> Yeah, good points. I would support moving nifi-registry-core/nifi-registry-docker to a top level module that is responsible for both dockerhub and dockermaven. Do you think we should do that now or save that improvement for later?


currently `nifi-registry/nifi-registry-core/nifi-registry-docker/dockerhub` is not a maven project (no pom.xml in its folder) this is another difference between the NiFi dockerhub and the NiFi Registry counter part. if we were to move it out from under the `nifi-registry-core` hierarchy, i'd also go for turning it into a maven project. 

If it is OK with you, i'd rather file a separate Jira / PR for that move & mavenization, to keep this PR small & single-work-item-focused.

kevdoran:
+1 for relocating it and making it a maven project, as well as for doing that as a separate jira/pr. I will re-review this PR as-is.

ekovacs:
> +1 for relocating it and making it a maven project, as well as for doing that as a separate jira/pr. I will re-review this PR as-is.

just created a follow up JIRA about it:
https://issues.apache.org/jira/browse/NIFIREG-342

PR #252, opened 2020-01-03
[NIFIREG-352] update frontend deps

scottyaslan:


alopresto:
Reviewing...

Simulalex:
Is this PR intended to address the issues found by `npm audit`? If so, I think something like this (https://github.com/apache/nifi-registry/compare/master...Simulalex:master) might address more of the findings.

If that's not what this PR is intended to fix, I can open a new ticket/PR for these changes.

PR #254, opened 2020-01-23
NIFIREG-355 Adding profiles to optionally include addtional hadoop li…

bbende:
…bs in ext/ranger/lib


mcgilman:
Thanks for the PR @bbende! This has been merged to master.

PR #255, opened 2020-01-23
NIFIREG-353 Add ShellUserGroupProvider and relax checks to allow a us…

bbende:
…er to have same identity as a group


joewitt:
+1 thanks Bryan.  can you please self merge

PR #256, opened 2020-01-31
NIFIREG-356 Fixing additional classpath resources for Authorizers and…

bbende:
… ensuring proper context ClassLoader is set

The original problem I was setting out to solve is what was described in the JIRA where didn't do anything with the classpathResources here:

https://github.com/apache/nifi-registry/blob/master/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/AuthorizerFactory.java#L340

It turned out there were a few other issues too...

- The introduction of the "FrameworkAuthorizer" broke some of the logic in the AuthorizerFactory that was trying to wrap the onConfigured call with the appropriate context ClassLoader

- The creation of the UGP, AP, and Authorizer should also set the context ClassLoader so it is set during the call to initialize, this is consistent with these methods on NiFi side

- The integrity checks and framework authorizer shouldn't be installed inside createAuthorizer, there could be multiple authorizers and it should only be done around the selected one


mcgilman:
Will review...

mcgilman:
Thanks for the PR @bbende! This have been merged to master.

PR #257, opened 2020-02-07
NIFIREG-358 Ensure full NiFiUser object is used when authorizing proxies

bbende:


bbende:
After going through the code some more, I am going to close this and take a different approach, will put up a new PR soon

PR #258, opened 2020-02-12
NIFIREG-358 Refactoring proxy authorization to be part of Authorizables

bbende:
- Remove framework authorizer
- Introduce ProxyChainAuthorizable to wrap all Authorizables
- Introduce PublicResourceAuthorizable to wrap bucket Authorizable


nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/resource/Authorizable.java
```diff
@@ -211,18 +205,10 @@ default void authorize(Authorizer authorizer, RequestAction action, NiFiUser use
             userContext = null;
         }
 
-        final List<String> proxyChain = new ArrayList<>();
-        NiFiUser proxyUser = user.getChain();
-        while (proxyUser  != null) {
-            proxyChain.add(proxyUser.getIdentity());
-            proxyUser = proxyUser.getChain();
-        }
```
bbende: Do we consider this a breaking change?

In 0.5.0 we added the proxyIdentities field and build methods to AuthorizationRequest. I left those there, but stopped populating them since we don't want to use them anymore. So it won't break compilation if some was using those, but they won't have values anymore. I can put this back if we want to leave it.

kevdoran:
Reviewing...

nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/resource/Authorizable.java
kevdoran: Good question. I would not consider this a breaking change given that the provider processing the authorization request will have to make an authorization decision based on the information that is passed in, and the responsibility for authorizing proxies has been moved so that it is no longer a concern for the provider impls.

PR #263, opened 2020-03-03
NIFIREG-366: updating pom parent version

ekovacs:
updating parent pom version to same as NiFi uses. 

associated ticket:
https://issues.apache.org/jira/browse/NIFIREG-366


ekovacs:
@bbende, could you please take a look?

pvillard31:
Merged to master, thanks @ekovacs

PR #279, opened 2020-05-19
[NIFIREG-389] remove npx and use npm-force-resolutions

scottyaslan:


apiri:
reviewing

apiri:
Ran into some test issues (https://issues.apache.org/jira/browse/NIFIREG-391) but verified this corrected build issues existing on master that now work in my Ubuntu environment.  Changes look good here!

PR #281, opened 2020-05-21
NIFIREG-394: call onConfigured with the same classloader

ChrisEnglert:


bbende:
@ChrisEnglert  thanks for submitting, sorry for the delay in looking at it...

Since we already have a precedent for how this done on the Authorizer's onConfigured [1] and we also have to do this for AccessPolicyProvider, in addition to UserGroupProvide, can we make them all consistent with the way the Authorizer is already done?

All three can be this with in the existing for loops without having to modify the create methods:
```
final ClassLoader instanceClassLoader = instance.getClass().getClassLoader();
try (final ExtensionCloseable extClosable = ExtensionCloseable.withClassLoader(instanceClassLoader)) {
    instance.onConfigured(loadAuthorizerConfiguration(provider.getIdentifier(), provider.getProperty()));
}
```

[1] https://github.com/apache/nifi-registry/blob/master/nifi-registry-core/nifi-registry-framework/src/main/java/org/apache/nifi/registry/security/authorization/AuthorizerFactory.java#L214-L217

PR #284, opened 2020-07-06
NIFIREG-385 Returning default Revision with version 0 on RevisableEnt…

bbende:
…ity when revisions are disabled, so that there is always a non-null revision in case it is referenced


markap14:
Good catch @bbende, changes look good to me. +1

PR #285, opened 2020-07-08
NIFIREG-400 Upgraded dependency versions.

alopresto:
Upgraded the jersey glassfish dependency version and ran a full build to verify.


bbende:
Looks good, merging

PR #286, opened 2020-07-09
NIFIREG-398 Change FlowFileConcurreny from enum to String to allow fo…

bbende:
…r flexible set of values from NiFi side


markap14:
Thanks @bbende all looks good. +1 merged to main

PR #292, opened 2020-08-04
NIFIREG-408 Adding clients for tenants and policies

bbende:
NOTE: This is includes the PR for NIFIREG-409 as that was required to make the integration test work correctly in this PR.

#### Description of PR

_Enables X functionality; fixes bug NIFIREG-YYYY._

### For all changes:
- [ ] Is there a JIRA ticket associated with this PR? Is it referenced 
     in the commit message?

- [ ] Does your PR title start with **NIFIREG-XXXX** where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [ ] Has your PR been rebased against the latest commit within the target branch (typically `main`)?

### For code changes:
- [ ] Have you verified that the full build is successful on JDK 8?

### Note:

### DIFF
--- a/nifi-registry-core/nifi-registry-client/src/main/java/org/apache/nifi/registry/client/NiFiRegistryClient.java
+++ b/nifi-registry-core/nifi-registry-client/src/main/java/org/apache/nifi/registry/client/NiFiRegistryClient.java
@@ -33,6 +33,8 @@ public interface NiFiRegistryClient extends Closeable {
      */
     BucketClient getBucketClient(String ... proxiedEntity);
 
+    //-------------------------------------------------------------------------------------------
+
     /**
      * @return the client for interacting with flows
      */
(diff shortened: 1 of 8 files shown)


kevdoran:
Thanks @bbende. I started reviewing this. Locally I rebased on `main` after merging #291. 

I'm not sure why, but although the integration tests are passing in GH Actions, they are failing for me locally

```
Results :


Tests in error:
  SecureNiFiRegistryClientIT.testTenantsClient:290 » SSL Connection has been shu...
```

Looking into it, but wanted to make you aware and ask if you had seen this?

bbende:
Thanks for reviewing!  Hmmm I do remember seeing this error once, but I think I only encountered it when running the IT test from Intellij and then after fixing the other revision problem I never saw it again, but interesting that it is happening from a maven build here.

bbende:
I can't figure out why that error seems to happen only on the updateUserGroup call, but I tried a few things and so far switching the order in the test where it does "create, retrieve, update, delete", instead of "create, update, retrieve, delete", seems to make it consistently pass. 

I feel like this is more some testing environmental issue, since even in the previous scenario, the create and update were working on the server side, but for some reason the connection was closed by the time the client tried to read the response.

kevdoran:
Hey, @bbende. I agree that it is a testing issue, and also that this change, although it doe not fix the root cause, does make the test seemingly pass reliably. +1 from me, will merge. We can monitor the testing setup to see if we need to fi the root cause at some point... thanks!

PR #297, opened 2020-08-24
NIFIREG-413 Add test DataSourceFactory for Postgress 11 and upgrade F…

bbende:
…lyway

Run tests with:
`mvn verify -Dspring.profiles.active=postgres-11`


pvillard31:
Is there any prerequisite to run the tests with Docker? I get errors with the below stacktraces:
> java.lang.IllegalStateException: Failed to load ApplicationContext
> ...
> Caused by: java.lang.IllegalStateException: Unable to load cache item
> ...
> Caused by: java.lang.ExceptionInInitializerError: null
> ...
> Caused by: java.lang.IllegalStateException: Could not find a valid Docker environment. Please see logs and check configuration

bbende:
You need to have a Docker daemon running locally. I have Docker Desktop for Mac.

pvillard31:
OK I was convinced that Docker was running but it was not... everything is looking good, merging to main, thanks @bbende

pvillard31 (approved):

PR #319, opened 2021-04-19
NIFIREG-395 - Implemented the ability to import and export versioned flows through the UI

mtien-apache:

#### Description of PR

- [WIP] This PR adds the ability for a user to take a previously saved versionedFlowSnapshot JSON file and import it as a new flow definition to an existing bucket or import it as the next new version of an existing flow through the UI. 

- This PR also adds the ability to export a versionedFlowSnapshot as a JSON file, which removes snapshot metadata associated with the given Registry. 

- New dialogs have been created to handle import and export.

- The REST endpoints added in `BucketFlowResource` are - 
     - `importFlow` - `/buckets/{bucketId}/flows/import`
     - `importVersionedFlow` - `/buckets/{bucketId}/flows/{flowId}/versions/import`
     - `exportVersionedFlow`- `/buckets/{bucketId}/flows/{flowId}/versions/{versionNumber: \d+}/export`

### For all changes:
- [x] Is there a JIRA ticket associated with this PR? Is it referenced 
     in the commit message?

- [x] Does your PR title start with **NIFIREG-XXXX** where XXXX is the JIRA number you are trying to resolve? Pay particular attention to the hyphen "-" character.

- [x] Has your PR been rebased against the latest commit within the target branch (typically `main`)?

### For code changes:
- [x] Have you verified that the full build is successful on JDK 8?

### Note:

### DIFF
--- a/nifi-registry-core/nifi-registry-web-api/pom.xml
+++ b/nifi-registry-core/nifi-registry-web-api/pom.xml
@@ -247,6 +247,15 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.rat</groupId>
+                <artifactId>apache-rat-plugin</artifactId>
+                <configuration>
+                    <excludes combine.children="append">
+                        <exclude>src/test/resources/test-versioned-flow-snapshot.json</exclude>
+                    </excludes>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 
@@ -422,10 +431,6 @@
             <groupId>io.jsonwebtoken</groupId>
             <artifactId>jjwt</artifactId>
             <version>0.7.0</version>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
         </dependency>
         <!-- Test Dependencies -->
         <dependency>
--- a/nifi-registry-core/nifi-registry-web-api/pom.xml
+++ b/nifi-registry-core/nifi-registry-web-api/pom.xml
@@ -427,12 +427,6 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
-        <dependency>
-            <groupId>com.google.protobuf</groupId>
-            <artifactId>protobuf-java</artifactId>
-            <version>3.11.4</version>
-            <scope>compile</scope>
-        </dependency>
         <!-- Test Dependencies -->
         <dependency>
             <groupId>org.apache.nifi.registry</groupId>
--- a/nifi-registry-core/nifi-registry-web-api/pom.xml
+++ b/nifi-registry-core/nifi-registry-web-api/pom.xml
@@ -423,6 +423,16 @@
             <artifactId>jjwt</artifactId>
             <version>0.7.0</version>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
+            <version>3.11.4</version>
+            <scope>compile</scope>
+        </dependency>
         <!-- Test Dependencies -->
         <dependency>
             <groupId>org.apache.nifi.registry</groupId>
--- a/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
+++ b/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
@@ -16,7 +16,6 @@
  */
 package org.apache.nifi.registry.web.api;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
@@ -25,7 +24,6 @@ import io.swagger.annotations.ApiResponses;
 import io.swagger.annotations.Authorization;
 import io.swagger.annotations.Extension;
 import io.swagger.annotations.ExtensionProperty;
-import java.io.IOException;
 import javax.ws.rs.HeaderParam;
 import javax.ws.rs.core.HttpHeaders;
 import org.apache.commons.lang3.StringUtils;
@@ -33,6 +31,7 @@ import org.apache.nifi.registry.bucket.BucketItem;
 import org.apache.nifi.registry.diff.VersionedFlowDifference;
 import org.apache.nifi.registry.event.EventFactory;
 import org.apache.nifi.registry.event.EventService;
+import org.apache.nifi.registry.web.service.ExportedVersionedFlowSnapshot;
 import org.apache.nifi.registry.flow.VersionedFlow;
 import org.apache.nifi.registry.flow.VersionedFlowSnapshot;
 import org.apache.nifi.registry.flow.VersionedFlowSnapshotMetadata;
@@ -40,7 +39,6 @@ import org.apache.nifi.registry.revision.entity.RevisionInfo;
 import org.apache.nifi.registry.revision.web.ClientIdParameter;
 import org.apache.nifi.registry.revision.web.LongParameter;
 import org.apache.nifi.registry.security.authorization.user.NiFiUserUtils;
-import org.apache.nifi.registry.serialization.jackson.ObjectMapperProvider;
 import org.apache.nifi.registry.web.service.ServiceFacade;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
@@ -75,8 +73,6 @@ public class BucketFlowResource extends ApplicationResource {
     public BucketFlowResource(final ServiceFacade serviceFacade, final EventService eventService) {
         super(serviceFacade, eventService);
     }
-
-    private static final ObjectMapper OBJECT_MAPPER = ObjectMapperProvider.getMapper();
 
     @POST
     @Consumes(MediaType.APPLICATION_JSON)
@@ -455,29 +451,18 @@ public class BucketFlowResource extends ApplicationResource {
             @PathParam("versionNumber")
             @ApiParam("The version number") final Integer versionNumber) {
 
-        if (StringUtils.isBlank(bucketId)) {
-            throw new IllegalArgumentException("The bucket identifier is required.");
-        }
+        final ExportedVersionedFlowSnapshot exportedVersionedFlowSnapshot = serviceFacade.exportFlowSnapshot(bucketId, flowId, versionNumber);
 
-        if (StringUtils.isBlank(flowId)) {
-            throw new IllegalArgumentException("The flow identifier is required.");
-        }
+        final VersionedFlowSnapshot versionedFlowSnapshot = exportedVersionedFlowSnapshot.getVersionedFlowSnapshot();
 
-        if (versionNumber == null) {
-            throw new IllegalArgumentException("The version number is required.");
-        }
+        final String contentDisposition = String.format(
+                "attachment; filename=\"%s\"",
+                exportedVersionedFlowSnapshot.getFilename());
 
-        final VersionedFlowSnapshot versionedFlowSnapshot = serviceFacade.exportFlowSnapshot(bucketId, flowId, versionNumber);
-
-        final String versionedFlowSnapshotJsonString = serializeToJson(versionedFlowSnapshot);
-
-        final String flowName = versionedFlowSnapshot.getFlowContents().getName();
-        final String dashFlowName = flowName.replaceAll("\\s", "-");
-        final String filename = String.format("%s-version-%d.json", dashFlowName, versionedFlowSnapshot.getSnapshotMetadata().getVersion());
-
-        final String contentDisposition = String.format("attachment; filename=\"%s\"", filename);
-
-        return generateOkResponse(versionedFlowSnapshotJsonString).header(HttpHeaders.CONTENT_DISPOSITION, contentDisposition).build();
+        return generateOkResponse(versionedFlowSnapshot)
+                .header(HttpHeaders.CONTENT_DISPOSITION, contentDisposition)
+                .header("Filename", exportedVersionedFlowSnapshot.getFilename())
+                .build();
     }
 
     @GET
@@ -633,13 +618,5 @@ public class BucketFlowResource extends ApplicationResource {
         }
 
         flowSnapshot.setSnapshotMetadata(metadata);
-    }
-
-    private String serializeToJson(final VersionedFlowSnapshot dto) {
-        try {
-            return OBJECT_MAPPER.writeValueAsString(dto);
-        } catch (IOException e) {
-            throw new IllegalArgumentException("Deserialization of selected flow failed", e);
-        }
     }
 }
--- a/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
+++ b/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
@@ -558,7 +558,11 @@ public class BucketFlowResource extends ApplicationResource {
 
         final String versionedFlowSnapshotJsonString = serializeToJson(versionedFlowSnapshot);
 
-        final String contentDisposition = String.format("attachment; filename=\"flow-version-%d.json\"", versionNumber);
+        final String flowName = versionedFlowSnapshot.getFlowContents().getName();
+        final String dashFlowName = flowName.replaceAll("\\s", "-");
+        final String filename = String.format("%s-version-%d.json", dashFlowName, versionedFlowSnapshot.getSnapshotMetadata().getVersion());
+
+        final String contentDisposition = String.format("attachment; filename=\"%s\"", filename);
 
         return generateOkResponse(versionedFlowSnapshotJsonString).header(HttpHeaders.CONTENT_DISPOSITION, contentDisposition).build();
     }
--- a/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
+++ b/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
@@ -299,8 +299,8 @@ public class BucketFlowResource extends ApplicationResource {
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(
-            value = "Upload flow version",
-            notes = "Uploads the next version of a flow. The version number of the object being created must be the " +
+            value = "Import flow version",
+            notes = "Import the next version of a flow. The version number of the object being created will be the " +
                     "next available version integer. Flow versions are immutable after they are created.",
             response = VersionedFlowSnapshot.class,
             extensions = {
@@ -323,7 +323,7 @@ public class BucketFlowResource extends ApplicationResource {
             @ApiParam(value = "The flow identifier")
             final String flowId,
             @ApiParam("file") final VersionedFlowSnapshot versionedFlowSnapshot,
-            @HeaderParam("comments") final String comments) {
+            @HeaderParam("Comments") final String comments) {
 
         final VersionedFlowSnapshot createdSnapshot = serviceFacade.importVersionedFlowSnapshot(versionedFlowSnapshot, bucketId, flowId, comments);
         publish(EventFactory.flowVersionCreated(createdSnapshot));
--- a/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
+++ b/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
@@ -24,6 +24,7 @@ import io.swagger.annotations.ApiResponses;
 import io.swagger.annotations.Authorization;
 import io.swagger.annotations.Extension;
 import io.swagger.annotations.ExtensionProperty;
+import java.net.URI;
 import javax.ws.rs.HeaderParam;
 import javax.ws.rs.core.HttpHeaders;
 import org.apache.commons.lang3.StringUtils;
@@ -310,6 +311,7 @@ public class BucketFlowResource extends ApplicationResource {
             }
     )
     @ApiResponses({
+            @ApiResponse(code = 201, message = HttpStatusMessages.MESSAGE_201),
             @ApiResponse(code = 400, message = HttpStatusMessages.MESSAGE_400),
             @ApiResponse(code = 401, message = HttpStatusMessages.MESSAGE_401),
             @ApiResponse(code = 403, message = HttpStatusMessages.MESSAGE_403),
@@ -327,7 +329,8 @@ public class BucketFlowResource extends ApplicationResource {
 
         final VersionedFlowSnapshot createdSnapshot = serviceFacade.importVersionedFlowSnapshot(versionedFlowSnapshot, bucketId, flowId, comments);
         publish(EventFactory.flowVersionCreated(createdSnapshot));
-        return Response.status(Response.Status.CREATED).entity(createdSnapshot).build();
+        String locationUri = createdSnapshot.getSnapshotMetadata().getLink().getUri().getPath();
+        return generateCreatedResponse(URI.create(locationUri), createdSnapshot).build();
     }
 
     @GET
--- a/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
+++ b/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
@@ -16,6 +16,7 @@
  */
 package org.apache.nifi.registry.web.api;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
@@ -24,6 +25,10 @@ import io.swagger.annotations.ApiResponses;
 import io.swagger.annotations.Authorization;
 import io.swagger.annotations.Extension;
 import io.swagger.annotations.ExtensionProperty;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import javax.ws.rs.core.HttpHeaders;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.nifi.registry.bucket.BucketItem;
 import org.apache.nifi.registry.diff.VersionedFlowDifference;
@@ -36,7 +41,9 @@ import org.apache.nifi.registry.revision.entity.RevisionInfo;
 import org.apache.nifi.registry.revision.web.ClientIdParameter;
 import org.apache.nifi.registry.revision.web.LongParameter;
 import org.apache.nifi.registry.security.authorization.user.NiFiUserUtils;
+import org.apache.nifi.registry.serialization.jackson.ObjectMapperProvider;
 import org.apache.nifi.registry.web.service.ServiceFacade;
+import org.glassfish.jersey.media.multipart.FormDataParam;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
@@ -56,6 +63,7 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import java.util.List;
 import java.util.SortedSet;
+import org.springframework.web.util.UriUtils;
 
 @Component
 @Path("/buckets/{bucketId}/flows")
@@ -66,10 +74,14 @@ import java.util.SortedSet;
 )
 public class BucketFlowResource extends ApplicationResource {
 
+    public static final String INVALID_JSON_MESSAGE = "Deserialization of uploaded JSON failed";
+
     @Autowired
     public BucketFlowResource(final ServiceFacade serviceFacade, final EventService eventService) {
         super(serviceFacade, eventService);
     }
+
+    private static final ObjectMapper OBJECT_MAPPER = ObjectMapperProvider.getMapper();
 
     @POST
     @Consumes(MediaType.APPLICATION_JSON)
@@ -99,8 +111,8 @@ public class BucketFlowResource extends ApplicationResource {
 
         verifyPathParamsMatchBody(bucketId, flow);
 
-        final VersionedFlow createdFlow = serviceFacade.createFlow(bucketId, flow);
-        publish(EventFactory.flowCreated(createdFlow));
+        final VersionedFlow createdFlow = createAndPublishVersionedFlow(bucketId, flow);
+
         return Response.status(Response.Status.OK).entity(createdFlow).build();
     }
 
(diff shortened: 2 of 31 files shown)


mtien-apache:
This PR is still WIP, but is open for initial review. I still need to add integration tests and resolve some styling issues, specifically with text alignment in the dialogs. 

@exceptionfactory @thenatog @bbende @pvillard31 Would you please review? 

@scottyaslan @sardell @akocsis @andras Would you please review the frontend code?

Thank you!

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -291,6 +303,129 @@ public Response createFlowVersion(
...
+            throw new IllegalArgumentException("The flow identifier is required.");
+        }
+
+        // deserialize InputStream to a VersionedFlowSnapshot
+        VersionedFlowSnapshot versionedFlowSnapshot;
+
+        versionedFlowSnapshot = deserializeVersionedFlowSnapshot(in);
+
+        // clear or set the necessary snapShotMetadata
+        if (versionedFlowSnapshot.getSnapshotMetadata() != null) {
```
bbende, line 351: I think we always want to clear these metadata values for the case where someone has json from another registry that was obtained before your new export existed. We probably also need to protect against the metadata being null because I think it is null coming from "Download Flow" in NiFi, so maybe always set to a new metadata and then apply anything on top of that?

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -385,6 +520,64 @@ public Response getLatestFlowVersionMetadata(
...
+
+        if (StringUtils.isBlank(flowId)) {
+            throw new IllegalArgumentException("The flow identifier is required.");
+        }
+
+        if (versionNumber == null) {
+            throw new IllegalArgumentException("The version number is required.");
+        }
+
+        final VersionedFlowSnapshot versionedFlowSnapshot = serviceFacade.getFlowSnapshot(bucketId, flowId, versionNumber);
```
bbende, line 562: This is kind of minor, but it might be nice to move most of this logic behind the service facade to a new method like exportFlowSnapshot(bucketId, flowId, versionNumber). We tried to make it so most of the logic happens behind the service facade, and the REST end-points don't do too much.

There is also an example of an existing file-download for extension bundles, which uses the StreamingContent/StreamingOutput, it is not as big a deal for this case since the JSON probably won't ever be too large, but in case you want to try and use this approach...

https://github.com/apache/nifi-registry/blob/main/nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/ExtensionRepoResource.java#L437-L445

bbende (commented):
@mtien-apache nice work, this is going to be very helpful!

I left a few inline comments, but here are some overall results from testing with a secured registry where I authenticated with LDAP (i.e. using JWTs)...

Using JSON obtained from NiFi "Download flow definition"
- Initial import worked correctly
- Trying to import a second version of the flow resulted in an error "Version must be greater than zero, or use -1 to indicate latest version"
- Trying to download a version of the flow popped up the file download dialog with the name "versionedFlow", after clicking save Chrome showed a download for "versionedFlow.txt" and said "Failed - Needs authorization"

Using JSON obtained from another NiFi Registry (i.e. it had the snapshot metadata populated from the other registry):
- Import New Flow produced an error "No applicable policies could be found. Contact the system administrator.", I think because it was trying to store it based on the previous metadata that was never cleared out

exceptionfactory (commented):
Thanks for the work on this feature @mtien-apache!  Noted a few questions, will run through another review when it is no longer in draft status.

nifi-registry-core/nifi-registry-web-api/pom.xml
```diff
@@ -423,6 +423,16 @@
             <artifactId>jjwt</artifactId>
             <version>0.7.0</version>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
```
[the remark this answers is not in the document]
mtien-apache: Removed - this was accidentally leftover from some previous work.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
mtien-apache: Good idea to set new metadata first - I revised this to create a new snapshotMetadata then set any necessary properties from there.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
mtien-apache: I refactored the some of this logic into the service facade.

nifi-registry-core/nifi-registry-web-api/pom.xml
```diff
@@ -423,6 +423,16 @@
...
+            <version>3.11.4</version>
+            <scope>compile</scope>
+        </dependency>
```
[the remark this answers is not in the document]
mtien-apache: Removed - see above comment.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -385,6 +520,64 @@ public Response getLatestFlowVersionMetadata(
         return Response.status(Response.Status.OK).entity(latest).build();
     }
 
+    @GET
+    @Path("{flowId}/versions/{versionNumber: \\d+}/export")
```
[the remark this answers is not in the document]
mtien-apache: I don't think the space is necessary, but decided to follow the format of the other endpoints.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -385,6 +520,64 @@ public Response getLatestFlowVersionMetadata(
...
+
+        String attachmentName = "flow-version-" + versionNumber;
+
+        UriUtils.encodePath(attachmentName, StandardCharsets.UTF_8);
+
+        String filename = attachmentName + ".json";
+
+        final String versionedFlowSnapshotJsonString = serializeToJson(versionedFlowSnapshot);
+
+        return generateOkResponse(versionedFlowSnapshotJsonString).header(HttpHeaders.CONTENT_DISPOSITION, String.format("attachment; filename=\"%s\"", filename)).build();
```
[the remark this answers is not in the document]
mtien-apache: Agreed that's easier to read. I've made the suggested changes. Thanks!

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -291,6 +303,129 @@ public Response createFlowVersion(
...
+
+        // set remaining snapshot metadata
+        final String userIdentity = NiFiUserUtils.getNiFiUserIdentity();
+        versionedFlowSnapshot.getSnapshotMetadata().setAuthor(userIdentity);
+        versionedFlowSnapshot.getSnapshotMetadata().setVersion(-1);
+
+        final VersionedFlowSnapshot createdSnapshot = serviceFacade.createFlowSnapshot(versionedFlowSnapshot);
+        publish(EventFactory.flowVersionCreated(createdSnapshot));
+
+        return Response.status(Response.Status.OK).entity(createdSnapshot).build();
```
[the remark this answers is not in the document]
mtien-apache: This endpoint has been refactored and no longer returns a response here.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -291,6 +303,129 @@ public Response createFlowVersion(
...
+
+        versionedFlowSnapshot = deserializeVersionedFlowSnapshot(in);
+
+        // clear or set the necessary snapShotMetadata
+        if (versionedFlowSnapshot.getSnapshotMetadata() != null) {
+            versionedFlowSnapshot.getSnapshotMetadata().setBucketIdentifier(null);
+            versionedFlowSnapshot.getSnapshotMetadata().setFlowIdentifier(null);
+            versionedFlowSnapshot.getSnapshotMetadata().setLink(null);
+            versionedFlowSnapshot.getSnapshotMetadata().setVersion(-1);
+            versionedFlowSnapshot.getSnapshotMetadata().setVersion(-1);
```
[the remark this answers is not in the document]
mtien-apache: Good idea - I've changed to use a static variable and removed the duplicate line of code.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -291,6 +303,129 @@ public Response createFlowVersion(
...
+
+        // deserialize InputStream and create new VersionedFlowSnapshot
+        final VersionedFlowSnapshot versionedFlowSnapshot = deserializeVersionedFlowSnapshot(in);
+
+        setSnaphotMetadataIfMissing(bucketId, createdFlow.getIdentifier(), versionedFlowSnapshot);
+
+        // set remaining snapshot metadata
+        final String userIdentity = NiFiUserUtils.getNiFiUserIdentity();
+        versionedFlowSnapshot.getSnapshotMetadata().setAuthor(userIdentity);
+        versionedFlowSnapshot.getSnapshotMetadata().setVersion(-1);
```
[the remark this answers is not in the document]
mtien-apache: Changed to use a static variable.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -385,6 +520,64 @@ public Response getLatestFlowVersionMetadata(
...
+
+        versionedFlowSnapshot.setFlow(null);
+        versionedFlowSnapshot.setBucket(null);
+        versionedFlowSnapshot.getSnapshotMetadata().setBucketIdentifier(null);
+        versionedFlowSnapshot.getSnapshotMetadata().setFlowIdentifier(null);
+        versionedFlowSnapshot.getSnapshotMetadata().setLink(null);
+
+        String attachmentName = "flow-version-" + versionNumber;
+
+        UriUtils.encodePath(attachmentName, StandardCharsets.UTF_8);
```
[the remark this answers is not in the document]
mtien-apache: I've removed the encoding.

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/services/nf-registry.api.js
```diff
@@ -75,6 +75,113 @@ NfRegistryApi.prototype = {
...
+     * Retrieves the specified versioned flow snapshot for an existing droplet the registry has stored.
+     *
+     * @param {string}  dropletUri      The uri of the droplet to request.
+     * @param {number}  versionNumber   The version of the flow to request.
+     * @returns {*}
+     */
+    exportDropletVersionedSnapshot: function (dropletUri, versionNumber) {
+        var self = this;
+        var url = '../nifi-registry-api/' + dropletUri;
+        url += '/versions/' + versionNumber + '/export';
```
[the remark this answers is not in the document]
mtien-apache: I don't think it's necessary to split up. I have changed to one line.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -291,6 +303,129 @@ public Response createFlowVersion(
...
+    }
+
+    @POST
+    @Path("import")
+    @Consumes(MediaType.MULTIPART_FORM_DATA)
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(
+            value = "Create flow",
+            notes = "Creates a flow in the given bucket. The flow id is created by the server and populated in the returned entity.",
+            response = VersionedFlow.class,
```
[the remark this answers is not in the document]
mtien-apache: Changed to VersionedFlowSnapshot.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -291,6 +303,129 @@ public Response createFlowVersion(
...
+                            @ExtensionProperty(name = "resource", value = "/buckets/{bucketId}")})
+            }
+    )
+    @ApiResponses({
+            @ApiResponse(code = 400, message = HttpStatusMessages.MESSAGE_400),
+            @ApiResponse(code = 401, message = HttpStatusMessages.MESSAGE_401),
+            @ApiResponse(code = 403, message = HttpStatusMessages.MESSAGE_403),
+            @ApiResponse(code = 404, message = HttpStatusMessages.MESSAGE_404),
+            @ApiResponse(code = 409, message = HttpStatusMessages.MESSAGE_409)})
+    public Response importFlow(
```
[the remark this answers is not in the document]
mtien-apache: My thought for creating this endpoint was to mainly handle the uploaded file and clear the snapshot metadata. It also made it a bit easier for the frontend to only make one request to the server. I have refactored some logic here to minimize duplicating code. It now returns `createFlowVersion` endpoint.

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/components/explorer/grid-list/dialogs/import-new-flow/nf-registry-import-new-flow.js
```diff
@@ -0,0 +1,182 @@
...
+NfRegistryImportNewFlow.prototype = {
+    constructor: NfRegistryImportNewFlow,
+
+    ngOnInit: function () {
+        // only show buckets the user can write
+        var self = this;
+        self.writableBuckets = self.filterWritableBuckets(self.buckets);
+    },
+
+    //TODO: get from the service instead
```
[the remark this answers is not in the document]
mtien-apache: My goal is to have this be like a utility method in Angular service (`nf-registry.service.js`). It would only show the buckets the user is authorized to write. I've been having issues injecting it into this import-new-flow component, so I'm in the middle of troubleshooting that.

mtien-apache:
I have pushed a commit that resolves the importing and exporting issues found by @bbende, as well as addressing other feedback found above this comment. 

Thanks for the first reviews, @bbende and @exceptionfactory !

andrewmlim:
This new functionality is great! It is going to be extremely useful.

Here are some thoughts after some UI/UX testing:

- I think it would be an improvement if the "Delete data flow" action simply said "Delete flow".  We refer to that type of asset in the Registry as "flow" and the documentation also refers to it as a "flow".   The dialog that appears after selecting the action also says "Delete flow" so this change would help with consistency.
- The title for the "Delete flow" dialog should be changed to "Delete Flow" for case consistency with the "Import New Version" and "Download Version" dialogs.
- When a flow version is downloaded the default name of the file is always flow-version-#.json (where # is the version #). I think it would be helpful if the name would default to the flow name.  So if I had a flow named "Kafka2HDFS" and wanted to download version 3 of it, the default name would be "Kafka2HDFS-version-3.json".

andrewmlim:
Additional thoughts:

- In the "Import New Flow" and "Import New Version" dialogs, change "Data Flow Name" to "Flow Name" for same reasons stated earlier.
- In the "Import New Version" dialog, consider changing "Change Log Comment" to "Version Comments" which is what is used in NiFi when saving a flow version.

version_comments

andrewmlim:
Testing how the feature does error handling and noticed the following:

When importing new flow, if I select a JSON file that is valid JSON, but not a valid flow definition file, after selecting import I get the error message "Cannot create versioned flow snapshot - flowContents must not be null".  After closing the dialog, it seems like nothing was imported, but if you refresh the view a new flow is imported with version 0.  (As an aside, if you select "Download version" on this corrupt flow, the UI seems to open a dialog on the left side of the window.)

bad_json

mtien-apache:
@andrewmlim Thanks for reviewing! 

- The UI changes make sense to me, especially if the names will match what's in the documentation and NiFi. 
@moranr Do you have any additional thoughts before I proceed?

- For the "DELETE" flow button color in the 'Delete Flow' dialog is what's currently in Registry:
> * I noticed that the color of the "DELETE" button in the "Delete Flow" dialog is a different color (more red) than the other dialog buttons (IMPORT and DOWNLOAD). This appears to be intentional. The "DELETE" button for deleting a bucket is the same red, so understandable if the color difference is to make it stand out since it is a unrecoverable action.

- I'm looking into the invalid flow definition file error now.

moranr:
Hey @mtien-apache - thanks for all your work on this. @andrewmlim 's comments all make sense to me and I would proceed with his recommendations. RE the Delete button color, yes the vibrant red is intentional as Drew observed. I believe this color is also used in other "delete" scenarios, so perhaps double-check to make sure color usage is consistent.

mtien-apache:
> Hey @mtien-apache - thanks for all your work on this. @andrewmlim 's comments all make sense to me and I would proceed with his recommendations. RE the Delete button color, yes the vibrant red is intentional as Drew observed. I believe this color is also used in other "delete" scenarios, so perhaps double-check to make sure color usage is consistent.

@moranr Thanks for confirming - will proceed with the UI changes. @andrewmlim Confirmed the Delete button color is the same as other Delete buttons used throughout Registry (eg, Delete Buckets).

mtien-apache:
Pushed 2 commits: 

- Handled the scenario when an invalid snapshot file is uploaded during Import New Flow @andrewmlim 
- Refactored client side API import methods
     - In the upload file methods, I removed `formData` and now passing the file as an API param. Also using HTTP headers to pass additional data
     - In `uploadFlow`, I'm now making multiple calls to the server that re-uses the existing `BucketFlowResource.createFlow` method
- Refactored server side import methods
     - Removed `BucketFlowResource.importFlow`
     - Refactored logic from `BucketFlowResource.importVersionedFlow` to the service facade
   
UI changes: 
- Fixed remaining UI style issues
- In the 'Import New Flow' dialog, the initial value for the Bucket dropdown menu is set to the Bucket the user is currently viewing. Otherwise, no value is set when viewing all buckets.
- Added messages to inform the user why the 'Import New Flow' button is displayed
     - One displays a message in the center of the page when there are no existing buckets 
     - Another message is displayed under the button buckets exist but the user does not have any write permissions

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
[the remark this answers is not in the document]
mtien-apache: @bbende I went ahead and removed the `importFlow` method. Now when importing a new flow, a call will first be made to `createFlow`, then another to `importVersionedFlow`.

andrewmlim:
Testing the latest changes and noticed that whenever I select "Import new version", the version is always shown as "v.2" in the "Import New Version" dialog even if is it the 3rd, 4th, 5th, etc. version that I am importing.

andrewmlim:
In the Actions drop-down, "Import new version" is always highlighted even when another selection has the focus.

Actions_highlight

andrewmlim:
> Related to my last comment, here are the messages in the error dialogs I have found so far:
> 
> * Non flow definition JSON file: "Cannot create versioned flow snapshot - flowContents must not be null"
> * Incomplete flow definition JSON file: "Deserialization of uploaded JSON failed"
> 
> It would be great if we could make these error messages more user friendly. But that could definitely be done as a separate Jira/PR. I wouldn't block this PR to address this.

For the incomplete flow definition JSON file, the error I see now when importing is: "Unexpected end-of-input: expected close marker for Object (start marker at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 1]) at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 2, column: 1]"

which is more specific to the issue (that there is no closing } in the JSON), but less readable.

mtien-apache:
> > Related to my last comment, here are the messages in the error dialogs I have found so far:
> > 
> > * Non flow definition JSON file: "Cannot create versioned flow snapshot - flowContents must not be null"
> > * Incomplete flow definition JSON file: "Deserialization of uploaded JSON failed"
> > 
> > It would be great if we could make these error messages more user friendly. But that could definitely be done as a separate Jira/PR. I wouldn't block this PR to address this.

@andrewmlim I refactored the resource method to let the Spring framework to automatically handle the deserialization and removed the "Deserialization of uploaded JSON failed" message. 

Agree to improve the error message in a separate PR for the "Cannot create versioned flow snapshot - flowContents must not be null" message.

> For the incomplete flow definition JSON file, the error I see now when importing is: "Unexpected end-of-input: expected close marker for Object (start marker at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 1]) at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 2, column: 1]"
> 
> which is more specific to the issue (that there is no closing } in the JSON), but less readable.

After investigating this exception message, it seems the exception is being caught by an `ExceptionMapper` and translating it to a BadRequest. It may be from `jackson-databind` in the Spring framework, which is not something we have much control over. It's possible to write a custom ExceptionMapper for a BadRequest, but the custom message could end up applying across the app for any method throwing that type of Exception, which can be addressed in a separate PR.

Although I agree the error message could be better, it at least points to which line the error is found. I believe this is sufficient for the time being. Thoughts?

andrewmlim:
> 
> Although I agree the error message could be better, it at least points to which line the error is found. I believe this is sufficient for the time being. Thoughts?

I agree that it is sufficient for now. All of the error messages can be handled outside of this PR. Thanks!

mtien-apache:
Pushed a commit:
- Added a new `ExportedVersionedFlowSnapshot` object
- Fixed Import New Version dialog to show the next version to be imported
- Refactored frontend and server side code.
- Added frontend and backend integration tests.

Will add more frontend tests.

mtien-apache:
> In the Actions drop-down, "Import new version" is always highlighted even when another selection has the focus.
> 
> Actions_highlight

@andrewmlim (cc: @moranr ) It seems like the initial focus behavior is based on keyboard actions and from an accessibility perspective, it's the expected behavior. When using the arrow keys, the focus moves with the new selection. I think what is making it confusing is that the keyboard and mouseover selections have the same red highlighting styles. I believe this highlighting behavior was always present but probably wasn't an issue since there had only been 1 menu item (Delete). 

One proposed fix is I can dig in to change the styles between a keyboard selection and mouseover selection, something like in other dropdown menus in Registry:

Screen Shot 2021-05-04 at 11 29 44 AM

Thoughts?

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -291,6 +294,42 @@ public Response createFlowVersion(
         return Response.status(Response.Status.OK).entity(createdSnapshot).build();
     }
 
+    @POST
+    @Path("{flowId}/versions/import")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(
+            value = "Upload flow version",
```
exceptionfactory, line 302: Should this be changed to Import flow version?
```suggestion
            value = "Import flow version",
```

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -291,6 +294,42 @@ public Response createFlowVersion(
...
+            notes = "Uploads the next version of a flow. The version number of the object being created must be the " +
+                    "next available version integer. Flow versions are immutable after they are created.",
```
exceptionfactory, line 304: The notes make it sound like the version number must be specified, as opposed to being generated.
```suggestion
            notes = "Import the next version of a flow. The version number of the object being created will be the " +
                    "next available version integer. Flow versions are immutable after they are created.",
```

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/service/StandardServiceFacade.java
```diff
@@ -360,6 +362,48 @@ public VersionedFlowSnapshot getLatestFlowSnapshot(final String flowIdentifier)
...
+                                                             String flowIdentifier, String comments) {
+        // set new snapshotMetadata
+        final VersionedFlowSnapshotMetadata metadata = new VersionedFlowSnapshotMetadata();
+        metadata.setBucketIdentifier(bucketIdentifier);
+        metadata.setFlowIdentifier(flowIdentifier);
+        metadata.setVersion(LATEST_VERSION);
+
+        // if there are new comments, then set it
+        // otherwise, keep the original comments
+        if (!StringUtils.isBlank(comments)) {
```
exceptionfactory, line 376: This could be changed to `StringUtils.isNotBlank()`
```suggestion
        if (StringUtils.isNotBlank(comments)) {
```

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/service/StandardServiceFacade.java
```diff
@@ -360,6 +362,48 @@ public VersionedFlowSnapshot getLatestFlowSnapshot(final String flowIdentifier)
...
+            metadata.setComments(comments);
+        } else if (versionedFlowSnapshot.getSnapshotMetadata() != null && versionedFlowSnapshot.getSnapshotMetadata().getComments() != null) {
```
exceptionfactory, line 378: Should this line check for blank comments?
```suggestion
        } else if (versionedFlowSnapshot.getSnapshotMetadata() != null && StringUtils.isNotBlank(versionedFlowSnapshot.getSnapshotMetadata().getComments())) {
```

nifi-registry-core/nifi-registry-web-api/src/test/java/org/apache/nifi/registry/web/api/FlowsIT.java
```diff
@@ -541,4 +544,171 @@ public void testFlowNameUniquePerBucket() throws Exception {
...
+        flowSnapshotMetadata.setBucketIdentifier(bucketId);
+        flowSnapshotMetadata.setFlowIdentifier(flowId);
+        flowSnapshotMetadata.setComments("This is a snapshot created by an integration test.");
+
+        // Create a VersionedFlowSnapshot
+        final VersionedFlowSnapshot flowSnapshot = new VersionedFlowSnapshot();
+        flowSnapshot.setSnapshotMetadata(flowSnapshotMetadata);
+        flowSnapshot.setFlowContents(new VersionedProcessGroup()); // an empty root process group
+        flowSnapshot.getFlowContents().setName("Test Flow name");
+        flowSnapshot.getSnapshotMetadata().setVersion(-1);
```
exceptionfactory, line 579: Recommend setting a private static final variable named LATEST_VERSION and reusing here and in the other test method.

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/components/explorer/grid-list/dialogs/import-new-flow/nf-registry-import-new-flow.js
```diff
@@ -0,0 +1,192 @@
...
+
+NfRegistryImportNewFlow.prototype = {
+    constructor: NfRegistryImportNewFlow,
+
+    ngOnInit: function () {
+        this.writableBuckets = this.filterWritableBuckets(this.buckets);
+
+        // if there's only 1 writable bucket, always set as the initial value in the bucket dropdown
+        // if opening the dialog from the explorer/grid-list, there is no active bucket
+        if (typeof this.activeBucket === 'undefined') {
```
exceptionfactory, line 62: Can this be simplified to check against `undefined` without `typeof`?
```suggestion
        if (this.activeBucket === undefined) {
```

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/services/nf-registry.api.js
```diff
@@ -75,6 +75,137 @@ NfRegistryApi.prototype = {
...
+        var url = '../nifi-registry-api/' + dropletUri + '/versions/' + versionNumber + '/export';
+        var options = {
+            headers: headers,
+            observe: 'response'
+        };
+
+        return this.http.get(url, options).pipe(
+            map(function (response) {
+                // export the VersionedFlowSnapshot by creating a hidden anchor element
+                var stringSnapshot = encodeURIComponent(JSON.stringify(response.body, null, 2));
```
exceptionfactory, line 96: Is it necessary to call `JSON.stringify()` or is it possible to just get the text content of response.body?  If `response.body` is already JSON, that means it went through an unnecessary serialization step, which would be great to avoid if possible.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -291,6 +294,42 @@ public Response createFlowVersion(
...
+            @ApiResponse(code = 409, message = HttpStatusMessages.MESSAGE_409) })
+    public Response importVersionedFlow(
+            @PathParam("bucketId")
+            @ApiParam("The bucket identifier")
+            final String bucketId,
+            @PathParam("flowId")
+            @ApiParam(value = "The flow identifier")
+            final String flowId,
+            @ApiParam("file") final VersionedFlowSnapshot versionedFlowSnapshot,
+            @HeaderParam("comments") final String comments) {
```
exceptionfactory, line 326: Recommend capitalizing this header to follow HTTP conventions:
```suggestion
            @HeaderParam("Comments") final String comments) {
```

exceptionfactory (commented):
The recent updates look good @mtien-apache!  I noted a few additional minor recommendations.

bbende:
Latest changes look good!

One minor comment... when downloading a version, the window that pops up has the latest version labeled as "Latest Version (v.#)" but then the other items in the list are just "Version #", seems a little inconsistent that one has "v." and the others don't. Should the latest one just say "Latest Version (#)" or maybe "Version # (latest)" ?

Also, while writing this, I started thinking about how we have generally been referring to this as import/export, and the buttons for import use the word "Import", but the for export we are calling it "Download Version". I'm not too hung up on this, but started wondering if that should be called "Export Version" to better correspond with "Import"?

andrewmlim:
> One proposed fix is I can dig in to change the styles between a keyboard selection and mouseover selection, something like in other dropdown menus in Registry:
> 
> Screen Shot 2021-05-04 at 11 29 44 AM
> 
> Thoughts?

That would be nice if you could change the styles. That would definitely help distinguish the two difference selections.

andrewmlim:
I tested a secure Registry with different authorizations given to users (on the users themselves and also on bucket policies). Auth looks good!

andrewmlim:
> One minor comment... when downloading a version, the window that pops up has the latest version labeled as "Latest Version (v.#)" but then the other items in the list are just "Version #", seems a little inconsistent that one has "v." and the others don't. Should the latest one just say "Latest Version (#)" or maybe "Version # (latest)" ?
> 
> Also, while writing this, I started thinking about how we have generally been referring to this as import/export, and the buttons for import use the word "Import", but the for export we are calling it "Download Version". I'm not too hung up on this, but started wondering if that should be called "Export Version" to better correspond with "Import"?

I think the most consistent label would be "Latest (Version #)".

"Export Version" is a good suggestion.  Like Bryan, I don't think that change is critical.

moranr:
> I think the most consistent label would be "Latest (Version #)".

@mtien-apache  I like that suggestion.

> "Export Version" is a good suggestion. Like Bryan, I don't think that change is critical.

Also agree here, but not critical. Typically you see Upload/Download or Import/Export used in together.

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/services/nf-registry.api.js
mtien-apache: @exceptionfactory Good catch! I can return a `responseType: 'text'` and omit calling `JSON.stringify()`. The file content will not be in pretty print format but that seems ok to me.

nifi-registry-core/nifi-registry-web-ui/src/main/webapp/components/explorer/grid-list/dialogs/download-versioned-flow/nf-registry-download-versioned-flow.js
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Download specified versioned flow snapshot.
+     */
+    downloadVersion: function () {
+        var self = this;
+        var version = this.selectedVersion;
+
+        this.nfRegistryApi.exportDropletVersionedSnapshot(this.droplet.link.href, version).subscribe(function (response) {
+            if (!response.status || response.status === 200) {
```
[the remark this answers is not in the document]
mtien-apache: @exceptionfactory A successful response has typically been just the response body returned. So, when the body is returned (a success) there will not be a status property. I can see how this could be confusing at first glance and without much context. This if-statement is used throughout the codebase, so I decided to follow the conventions.

mtien-apache:
Pushed commits that addresses all remaining changes and PR feedback. The PR is ready for a full review. 
@sardell Would you review the frontend changes?

Recent changes include:

- Changed the dialog title from 'Download Version' to 'Export Version' and refactored the component. @bbende @andrewmlim @moranr Thanks for the feedback and agreed 'Export' seems more consistent  

- Changed the 'Export Version' dialog dropdown menu item to "Latest (Version #)"
- Changed the Actions menu hover style to differentiate from the focus style
- Added frontend tests. Note - With the way Angular and Javascript is currently configured in Registry, I had issues writing certain tests for the new components. It's been captured in a JIRA ticket: [NIFIREG-454](https://issues.apache.org/jira/browse/NIFIREG-454). Once resolved, more tests can be added.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -291,6 +294,42 @@ public Response createFlowVersion(
...
+                    "next available version integer. Flow versions are immutable after they are created.",
+            response = VersionedFlowSnapshot.class,
+            extensions = {
+                    @Extension(name = "access-policy", properties = {
+                            @ExtensionProperty(name = "action", value = "write"),
+                            @ExtensionProperty(name = "resource", value = "/buckets/{bucketId}") })
+            }
+    )
+    @ApiResponses({
+            @ApiResponse(code = 400, message = HttpStatusMessages.MESSAGE_400),
```
exceptionfactory, line 313: Adding an `ApiResponse` annotation with a code of `201` would be helpful for completeness in the generated documentation.

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
```diff
@@ -291,6 +294,42 @@ public Response createFlowVersion(
...
+            final String bucketId,
+            @PathParam("flowId")
+            @ApiParam(value = "The flow identifier")
+            final String flowId,
+            @ApiParam("file") final VersionedFlowSnapshot versionedFlowSnapshot,
+            @HeaderParam("Comments") final String comments) {
+
+        final VersionedFlowSnapshot createdSnapshot = serviceFacade.importVersionedFlowSnapshot(versionedFlowSnapshot, bucketId, flowId, comments);
+        publish(EventFactory.flowVersionCreated(createdSnapshot));
+        return Response.status(Response.Status.CREATED).entity(createdSnapshot).build();
```
exceptionfactory, line 330: Best practice when returning an HTTP 201 includes adding a `Location` header with the URL of the new resource.  See `ApplicationResource.generateCreatedResponse()` as a helper method for returning the Location header with the created entity.

exceptionfactory (commented):
Thanks for working through the feedback details @mtien-apache!  This looks just about good to go, I noted one more detail around the HTTP 201 Created response.

andrewmlim:
> Recent changes include:
> 
> * Changed the dialog title from 'Download Version' to 'Export Version' and refactored the component. @bbende @andrewmlim @moranr Thanks for the feedback and agreed 'Export' seems more consistent
> * Changed the 'Export Version' dialog dropdown menu item to "Latest (Version #)"
> * Changed the Actions menu hover style to differentiate from the focus style

These look good! Thanks for making all of these improvements @mtien-apache !

exceptionfactory (approved):

nifi-registry-core/nifi-registry-web-api/src/main/java/org/apache/nifi/registry/web/api/BucketFlowResource.java
mtien-apache: @exceptionfactory I've added the header. Thanks!

sardell:
@mtien-apache I just finished looking at the UI code and testing your changes in the UI. Nice work. I'm a +1 (non-binding).

mtien-apache:
Closed in favor of https://github.com/apache/nifi/pull/5107

